PR #5, opened 2023-05-05
Final updates for V1 to work

zFernand0:
Note:
- The last thing to do is to add the `deployment.yml` once it's added (and tested) in the main branch


zFernand0:
Found two system tests issues with URIMAPs of type client on a specific LPAR.

PR #147, opened 2024-09-20
chore: update release config to get cli/sdk with a non prerelease

zFernand0:
**What It Does**

Only release plugin CLI and SDK as 6.0.0 to public zowe.jfrog

### DIFF
--- a/.github/release.config.js
+++ b/.github/release.config.js
@@ -12,9 +12,8 @@ module.exports = {
         },
         {
             name: "main",
-            channel: "next",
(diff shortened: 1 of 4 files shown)


t1m0thyj (dismissed):
LGTM, thanks @zFernand0! The 3.0.0 VSIX will still be published as a GH release artifact but we can re-upload it later 😋

t1m0thyj (approved):

awharn (approved):

Issue #33, opened 2023-05-19
Have option to create team config file


JillieBeanSim:
The extension recognizes and uses team config files, but if no team config file is in place and user tries to create profile from CICS view the user is presented with v1 profile creation steps and isn't offered the option to create team config file.

zFernand0:
I'm afraid this didn't seem to make it into the roadmap [during PI planning](https://github.com/zowe/community/blob/master/Project%20Management/PI%20Planning/24PI4%20Planning/PI%20Planning%20Preparation%20by%20Squad/Zowe%20Explorer%20Squad%20-%2024PI4%20Objectives.md), however, this seems like something we need to address as soon as practicable 🙏

davenice:
PR https://github.com/zowe/cics-for-zowe-client/pull/160 means that the `+` icon on the CICS view now redirects to the config file correctly.

Right-clicking on a profile and choosing update pops up the bad dialog.
Choosing delete does nothing.

Chin is working this to completion 👌

chinmdas:
With the fixes going in as part of PR https://github.com/zowe/cics-for-zowe-client/issues/33 , following will be experienced by users upon click of `+` icon:

- `zowe.config.json` is **missing** -> **New Team Config File is created**
- `zowe.config.json` file is **available**, but CICS Profile is **missing** -> **Edit Team Config File**
- `zowe.config.json` file is **available** + CICS Profile is **available**  -> **Edit Team Config File**

PR #166, opened 2024-12-13
Initializing new Team Configuration File

chinmdas:
Initializing new Team Configuration File and remove the obsolete web page,

**What It Does**

- No more obsolete web page when user tries to create a new CICS Profile. Instead a new global team config file will get initialized.

**How to Test**
- If NO global zowe.config.json file exists: Creates a New.
- If zowe.config.json file exists, but NO CICS Profile is present: Gives option to Edit.
- If zowe.config.json file exists and CICS Profile is available: Gives option to Edit.

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog

**Additional Comments**


chinmdas:
This PR fixes https://github.com/zowe/cics-for-zowe-client/issues/33.

AndrewTwydell (changes requested):
I get this popup when clicking "add" with no config file. We should change that text to something more meaningful.

image

AndrewTwydell (commented):
When clicking edit config file when there are no profiles in there, we get a horrible error
image

image

AndrewTwydell:
When you click the Add button on other ZE views and no config file exists, you get this menu
image

Perhaps we should do the same to be consistent, rather than going straight into the Global or Project quickpick??

Joe-Winchester (approved):

chinmdas:
> I get this popup when clicking "add" with no config file. We should change that text to something more meaningful.
> 
> image

Yes, Agreed..!!! What can be appropriate info message? Do you think this makes sense?
`Team Configuration file not found. Creating new Team Configuration file`

chinmdas:
> When clicking edit config file when there are no profiles in there, we get a horrible error image
> 
> image

Nice Testing @AndrewTwydell ..!!
The logic of picking up `filepath` for `showQuickPick` is very tightly coupled with profiles availability in `zowe.config.json` file. Since, you have delete all the profiles entry, extraction of `filepath` broke.
I am going to fix this. Thanks..!!

chinmdas:
> When you click the Add button on other ZE views and no config file exists, you get this menu image
> 
> Perhaps we should do the same to be consistent, rather than going straight into the Global or Project quickpick??

Agreed...!!! Will implement this...!!

chinmdas:
> Hi @AndrewTwydell ,
> 
> Where did you get a .vsix from to do the tests for the screen shots you've shown above ? I did a code review which looked good but I'd like to do a test of the PR code. Could checkout the branch and launch from VS code launch.json but wondering if there is a .vsix build created for the PR ?

Thanks @Joe-Winchester .... I see all the valid points raised by @AndrewTwydell . I would implement his suggestions and seek your approval again..!!

Joe-Winchester:
On the CICS profile I still see the pop-up action `Update Profile` that opens the Connection Details dialog.  
The connection details dialog should be removed so it's not accessible at all (maybe just yank all the code), and instead `Update Profile` should probably drop into the editor for the `zowe.config.json` file itself.
(Note, this testing was me checking out the branch and doing a debug launch so it's possible I have a false test)

image

chinmdas:
> On the CICS profile I still see the pop-up action `Update Profile` that opens the Connection Details dialog. The connection details dialog should be removed so it's not accessible at all (maybe just yank all the code), and instead `Update Profile` should probably drop into the editor for the `zowe.config.json` file itself. (Note, this testing was me checking out the branch and doing a debug launch so it's possible I have a false test, or also possible this fix is outside the scope of the issue this PR is fixing, and maybe another issue should be raised for removing the deprecated page for profile editing alltogether.
> 
> image

Okay.... Thank you bringing out this scenario @Joe-Winchester . I will fix this as well.

AndrewTwydell (changes requested):
Looking good @chinmdas, few items to consider 👍🏼

AndrewTwydell:
With the interest of trying to align with ZE's behaviour, there are a couple of questions to answer... 

There are currently 2 clicks required to update a profile (update -> edit) when that's the only option. ZE have 2 because they have manage -> edit. What are we aiming for? If we are eventually going to imitate them, should we also change to give the same manage profile menu?

Secondly, we no longer get the option to create a project config, only global. Same as above, are we trying to be consistent? Probably! 

Note: we shouldn't have the delete profile option anymore but that can be another PR to remove it 🙂

chinmdas:
> With the interest of trying to align with ZE's behaviour, there are a couple of questions to answer...
> 
> There are currently 2 clicks required to update a profile (update -> edit) when that's the only option. ZE have 2 because they have manage -> edit. What are we aiming for? If we are eventually going to imitate them, should we also change to give the same manage profile menu?
> 
> Secondly, we no longer get the option to create a project config, only global. Same as above, are we trying to be consistent? Probably!
> 
> Note: we shouldn't have the delete profile option anymore but that can be another PR to remove it 🙂

Hi @Joe-Winchester @zFernand0 @davenice ,
Can we have your expert advice on this?
Just to build context, following is what this PR handles :

- `zowe.config.json` is missing -> `New Team Config File is created `-> _New Global Team Configuration file is created_
- `zowe.config.json` file is available, but CICS Profile is missing ->` Edit Team Config File` -> _Available Global Team Configuration file is opened for modification_
- `zowe.config.json` file is available + CICS Profile is available -> Edit Team Config File ->  _Available Global Team Configuration file is opened for modification_
- `Right click on CICS profile` -> `Update Profile` ->  _Available Global Team Configuration file is opened for modification_

So, The gaps that @AndrewTwydell has mentioned, should we do it as part of this PR or should we raise a new PR for that?

packages/vsce/src/trees/CICSTree.ts
```diff
@@ -112,24 +104,21 @@ export class CICSTree implements TreeDataProvider<CICSSessionTree> {
...
-              await this.createNewProfile();
-            }
+          if (profileNameToLoad.label.includes("\u270F")) {
+            // get all profiles of all types including zosmf
+            const profiles = configInstance.getAllProfiles();
+            const currentProfile =
+                profiles.length > 0 ? await ProfileManagement.getProfilesCache().getProfileFromConfig(profiles[0].profName) : null;
+            const filePath =
+                  currentProfile === null
+                    ? configInstance.getTeamConfig().opts.homeDir + "/zowe.config.json"
```
davenice, line 119: did you notice if there was a constant in Zowe Explorer we could use for zowe.confgi.json? or perhaps even for the whole profile path? that would be nice.

davenice: Is this something we can use? Or not the right thing?
https://github.com/zowe/zowe-explorer-vscode/blob/aea6d92717f4b765088a8d17e36e0c2c1a38a8d4/packages/zowe-explorer/src/extending/ZoweExplorerExtender.ts#L99

davenice (commented):
Looks good and seems good testing your branch out.

One question around whether we can use the Zowe Explorer code to get the path/filename 🤔

packages/vsce/src/trees/CICSTree.ts
davenice: Actually I reckon that is not that useful, it needs us to give the basic path and it just adds the filename. Probably not worth using.

packages/vsce/src/trees/CICSTree.ts
```diff
@@ -112,24 +104,21 @@ export class CICSTree implements TreeDataProvider<CICSSessionTree> {
...
+                    : (currentProfile?.profLoc.osLoc?.[0] ?? "");
+            await openConfigFile(filePath);
```
AndrewTwydell, line 121: It looks like filePath could be set to ""... Have you tested what happens when you call openConfigPath with an empty string? Might be worth reworking that so if can only ever be called with a valid parameter.

chinmdas: Resolved..!!!

packages/vsce/src/trees/CICSTree.ts
chinmdas: Resolving this thread..!!

packages/vsce/src/trees/CICSTree.ts
```diff
@@ -583,35 +557,21 @@ export class CICSTree implements TreeDataProvider<CICSSessionTree> {
...
-      }
+  async updateSessionHelper(profile?: imperative.IProfAttrs) {
+    const response = await window.showQuickPick([{ label: "Edit CICS Profile" }], {
+      ignoreFocusOut: true,
+      placeHolder: "Create a New Team Configuration File",
     });
+    if (response) {
+      const filePath = profile.profLoc.osLoc?.[0];
+      await openConfigFile(filePath);
+    }
```
AndrewTwydell, line 574: Sorry to be picky but, again, if osLoc is not set (which it doesn't have to be, it can be undefined), filePath can be set to undefined. Could do something similar to this, although you'd need to test and double check it, to only call openConfigFile with a valid input.

```suggestion
    if (response && profile.profLoc.osLoc?.[0]) {
      await openConfigFile(profile.profLoc.osLoc.[0]);
    } else {
      window.showErrorMessage("Some reasonable message about what you're trying to do??");
    }
```

packages/vsce/src/trees/CICSTree.ts
```diff
@@ -117,7 +117,7 @@ export class CICSTree implements TreeDataProvider<CICSSessionTree> {
             const filePath =
                   currentProfile === null
                     ? configInstance.getTeamConfig().opts.homeDir + "/zowe.config.json"
-                    : (currentProfile?.profLoc.osLoc?.[0] ?? "");
+                    : (currentProfile.profLoc.osLoc?.[0]);
```
davenice, line 120: just been discussing this with @chinmdas - we've decided that if the profile is not null, we should have profLoc and an entry in osLoc, therefore if we don't that's a problem and we should just fail here rather than in openConfigFile.

so we decided to remove the question mark.

packages/vsce/src/trees/CICSTree.ts
chinmdas: Actually, that condition you have added is redundant.
Update profile scenario will only be invoked on a available profile. So, If profile is available >> the Team Configuration file is available >> no need to check on` profile.profLoc.osLoc?.[0]`

AndrewTwydell:
The behaviour and flow of the profile experience is way better now @chinmdas, good work, I'm able to get to the config file when I need to and it handles odd edge cases nicely. I think it's a case of adding in those null checks that we've spoken about and the issue is resolved. We can create new issues for the problems we've found along the way 😉

Oh and are you able to squash your commits? Just to keep things as tidy as we can 😎

packages/vsce/src/trees/CICSTree.ts
chinmdas: I have resolved this. We are falling back to Team Config File Path if osLoc fails to give us profile location path. This will cover us from all the angles.

packages/vsce/src/trees/CICSTree.ts
chinmdas: I have resolved this. We are falling back to Team Config File Path if `osLoc` fails to give us profile location path. This will cover us from all the angles.

packages/vsce/src/trees/CICSTree.ts
```diff
@@ -583,35 +555,23 @@ export class CICSTree implements TreeDataProvider<CICSSessionTree> {
...
-        const profile = await ProfileManagement.updateProfile(message);
-        const position = this.loadedProfiles.indexOf(session);
-        await ProfileManagement.profilesCacheRefresh();
-        const updatedProfile = await ProfileManagement.getProfilesCache().loadNamedProfile(profile.profile.name, "cics");
-        await this.removeSession(session, updatedProfile, position);
-      } catch (error) {
-        // @ts-ignore
-        window.showErrorMessage(error);
-      }
+  async updateSessionHelper(profile?: imperative.IProfAttrs) {
```
AndrewTwydell, line 564: A profile is always provided to this method right? Therefore we don't need an optional parameter here.

```suggestion
  async updateSessionHelper(profile: imperative.IProfAttrs) {
```

davenice:
I think this looks pretty good - can you squash the commits down to 1 because it's just one piece of work?

That'll fix the DCO issue and make sure that whenever it gets merged it looks like one seamless piece? :-D

chinmdas:
> enever it gets merged it looks like one seamless piece

I dont know, if I get 2 approvals, I can check if I have an option of squash merge.

zFernand0:
no need to squash commit now 🙏 

You should be able to follow the preferred approach here:
- https://github.com/zowe/cics-for-zowe-client/pull/166/checks?check_run_id=34609378367

image

AndrewTwydell:
@zFernand0 do you have squash and merge enabled on the repo, it would be nice to keep the history as tidy as possible right? 🫣 Avoiding a load of main merges and revert commits is preferable?

Joe-Winchester:
> @zFernand0 do you have squash and merge enabled on the repo, it would be nice to keep the history as tidy as possible right? 🫣 Avoiding a load of main merges and revert commits is preferable?

Just checked the settings and it looks like it's enabled

image

davenice (approved):
Excellent. My preference is to organise commits before they're merged, simply because it's too easy for the person merging to click the wrong button and then my stream of commit-consciousness is listed in the history forever. Sadly I've been on both sides of that storyline!

But it sounds like we've got that covered today with the button :-) 👍

chinmdas:
Hi @Joe-Winchester @zFernand0 , Do I have merge access?

AndrewTwydell (approved):
LGTM - Thanks @chinmdas! 

I don't think we have the powers to merge to main, no - potentially for the best for the time being!

JillieBeanSim:
Hey @Joe-Winchester can add admin access to the team members you would like to be able to merge PRs 😸 
Should be able to do this under the Settings tab then setup group/users within the Access>Collaborators & Teams

zFernand0:
As it stands right now, only these two teams have access to merge PRs to protected branches on this monorepo
- https://github.com/orgs/zowe/teams/zowe-cics-explorer-administrators (Maintainer access)
- https://github.com/orgs/zowe/teams/zowe-cli-administrators (Admin access)

As for adding more people to the mainteners/admins group(s), that's always up to the squad (it doesn't matter how new someone might be) 😋 

If y'all want to add a few more folks to the CICS Administrators, feel free to bring it up on the next standup 😋 

---

As for squashing vs merging...
I'm personally a fan of squashing PRs, however We may need to so some testing with the deployment workflow before we decide to move to a squash-approach 😋 

Thanks Joe for looking into the squash vs merge configuration.

zFernand0:
Hey @chinmdas,
Did this commit resolve the problem DCO problem? c01d2b68d53d6b00ac36aa1ec13aabdbaf29072d 🥳 
Or did someone manually clicked the `Set DCO to Pass`? 🤔

zFernand0 (approved):
I really like the new flow 🙏 

As a follow up PR, we could have a Create CICS profile which just run the `zowe.all.config.init`with a warning to make people aware that new profiles will be created in their existing zowe.config.json file 😋 

(Similar to the Zowe Explorer flow) 😋

zFernand0:
Labeling: `release-minor` to include these enhancements from the SDK
image

chinmdas:
> Hey @chinmdas, Did this commit resolve the problem DCO problem? [c01d2b6](https://github.com/zowe/cics-for-zowe-client/commit/c01d2b68d53d6b00ac36aa1ec13aabdbaf29072d) 🥳 Or did someone manually clicked the `Set DCO to Pass` button? 🤔

Hi @zFernand0 ,
That commit did not pass DCO. I manually clicked `Set DCO to Pass` to check what option that gives me in further steps, but little I knew that one click will set the DCO to pass.

chinmdas:
> I really like the new flow 🙏
> 
> As a follow up PR, we could have a Create CICS profile which just run the `zowe.all.config.init`with a warning to make people aware that new profiles will be created in their existing zowe.config.json file 😋
> 
> (Similar to the Zowe Explorer flow) 😋

Sure, We will do this.
Along with that, there were other scenario's that @AndrewTwydell and @davenice spotted which needs improvisation in order to stay in sync with ZE Flow. I will be creating new change set for that.
Thanks for your promptness @zFernand0 ..!!

zFernand0:
> > Hey `@chinmdas`, Did this commit resolve the problem DCO problem? [c01d2b6](https://github.com/zowe/cics-for-zowe-client/commit/c01d2b68d53d6b00ac36aa1ec13aabdbaf29072d) 🥳 Or did someone manually clicked the `Set DCO to Pass` button? 🤔
> 
> Hi `@zFernand0` , That commit did not pass DCO. I manually clicked `Set DCO to Pass` to check what option that gives me in further steps, but little I knew that one click will set the DCO to pass.

Ideally, we should avoid clicking that button. There is just no configuration to remove that button. More details:
- https://github.com/dcoapp/app/issues/119

Issue #169, opened 2024-12-16
Create SDK method to retrieve CICS Result Cache


AndrewTwydell:
The getResource SDK method retrieves a set of results, which can include a cache token with #168. 

Having a way to use the cache token provided by the getResource method to retrieve cached or paginated results would allow us to fully migrate the VSCE to use the SDK rather than custom HTTP requests.

PR #171, opened 2024-12-17
Add CMCI query parameters to SDK getResource

AndrewTwydell:
**What It Does**
Allows a user to specify query parameters to be added to the CMCI query URI (SUMMONLY, NODISCARD, OVERRIDEWARNINGCOUNT) by adding an optional addition to the supplied params.

This will allow the VSCE to retrieve a cache token from CMCI which will enable pagination on large result sets.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

PR #172, opened 2024-12-17
Add SDK getCache method

AndrewTwydell:
Requires #171 - Merged!

**What It Does**
Resolves zowe/cics-for-zowe-client#169

Adds getCache method to the SDK to retrieve a cached result from CMCI using a cache token returned by getResource.

This will allow the VSCE to paginate large result sets using the SDK rather than axios HTTP requests.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the IBM® CICS® Plug-in for Zowe CLI will be documented 
 ## Recent Changes
 
 - Enhancement: Add optional query parameters on getResource SDK method. [#168](https://github.com/zowe/cics-for-zowe-client/issues/168)
+- Enhancement: Add getCache method to SDK. [#169](https://github.com/zowe/cics-for-zowe-client/issues/169)
 
 ## `6.1.0`
 
--- a/packages/sdk/__tests__/__unit__/cache/Cache.unit.test.ts
+++ b/packages/sdk/__tests__/__unit__/cache/Cache.unit.test.ts
@@ -1,0 +1,189 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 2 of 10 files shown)


davenice (approved):
Looks good I think.

zFernand0 (changes requested):
LGTM! 😋 
For these new set of APIs, (as well as SUMMONLY, NODISCARD, OVERRIDEWARNINGCOUNT), Could we add some system tests? 

Depending on people's environment, having system tests might help determine whether the APIs work as expected, or if there are edge cases to consider 😋 

---

We could handle system tests in a follow up PR (for both, cache, and the new query params) if you'd like to get this one merged sooner rather than later 🙏

zFernand0 (approved):
The functionality works great!
I like the pagination use case.

I was playing around, but I could not properly test the `OVERRIDEWARNINGCOUNT` queryParm (likely because the `DEFAULTWARNCNT` was never specified during the WUI configuration) 😢 

I say "Let's keep moving forward"  and let's handle the system tests in a separate PR 🙏 

---

For those testing...
<details><summary>Feel free to use this script</summary> 

```typescript
/* Save this file as `test.mjs` and run `node test.mjs` */

import { ProfileInfo } from "@zowe/imperative";
import * as sdk from "@zowe/cics-for-zowe-sdk";

(async () => {
    const profInfo = new ProfileInfo("zowe");
    await profInfo.readProfilesFromDisk();
    const profAttrs = profInfo.getDefaultProfile("cics");
    const mergedArgs = profInfo.mergeArgsForProfile(profAttrs, { getSecureVals: true });
    const session = ProfileInfo.createSession(mergedArgs.knownArgs);

    let response;
    response = (await sdk.getResource(session, {name: "CICSCSDResource", queryParams: {summonly: true, nodiscard: true}}));
    const cacheToken = response.response.resultsummary.cachetoken;
    console.log("Cache token: ", cacheToken);

    // Use the above cache token and play around with the `startIndex` and `count` values
    response = (await sdk.getCache(session, {cacheToken, startIndex:1, count:100})).response.records.cicscsdresource.map(x => x.defname);
    console.log(response);
})().catch((err) => {
    console.error(err);
    process.exit(1);
});
```

</details>

Issue #14, opened 2024-02-01
Migrate VSCE-only commands into the SDK


zFernand0:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

There seems to be an entire set of APIs that were only implemented in the CICS VSCE repo.

**Describe your enhancement idea**

Now that this is a monorepo, we should:
- [ ] migrate those to the SDK
- [ ] expose them through the plug-in

PR #173, opened 2024-12-17
Fix CICS connections breaking other ZE sessions

### LINKED ISSUE #124: VSCE: Unable to list `SIT-parms` for a region with self-signed certs
**Describe the bug**

Having rejectUnauthorized in the CICS profile (or base profile) does not matter when you try to show the SIT Parameters for the region

Expected to get the results of the following CICS reosurce
- Endpoint: `CICSSystemParameter`
- Parmameters: `PARMSRCE(COMBINED) PARMTYPE(SIT)`

Resulted in the following error:
image

**Describe your environment**

ZE VSCE: v2.15.2
CICS VSCE: v2.3.5
SCS: Yes

**Additional context**

N/A

### LINKED ISSUE #159: VSCE: using the CICS extension with a different host breaks Zowe Explorer
**Describe the bug**

I have multiple hosts in play.
hosta is a z/OSMF host which is https and using a self-signed certificate
hostb hosts a CMCI connection - either https with self-signed certificate or http (no ssl)
I am specifying `"rejectUnauthorized": false` on both profiles.

Using the CICS extension breaks the Zowe Explorer extension.

**Recreation steps**

I connect to hosta with z/OSMF correctly and browse data sets.
I connect to hostb with the CICS extension correctly and browse CICS resources.
I attempt to refresh my hosta z/OSMF data sets and receive this error: `Retrieving response from MVS list API Error: Failed to send an HTTP request.`

I reload my VS Code and able to connect to z/OSMF on hosta correctly, again.

If I stick to just using hostb consistently, I am able to use both z/OSMF and CMCI.

**Describe your environment**

- Zowe CLI version installed (run command `zowe --version`): CLI Version: 8.8.2, Zowe Release Version: v3.0.0
- Zowe Explorer version installed: 3.0.3
- Installed plug-ins and their version numbers (run command `zowe plugins list --short`): no plugins
- Installed z/OS-related VSCode Extensions and their version numbers: CICS - 3.0.1
- Node.js and NPM versions installed (run `node --version`, `npm --version`): v22.9.0, 10.8.3

AndrewTwydell:
**What It Does**
- Changes method of pulling data from CMCI from Axios HTTP requests to the CICS SDK's getResource and getCache methods
  - Removes makeRequest method - refactored response handling
  - Refactored getPlexInfo method into smaller methods depending on what's provided in the profile
- Removes the global rejectUnauthorized overrides which broke the other ZE sessions
- Created toArray utility to replace all cases where we had to do a Array.isArray check

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Resolves #159
Resolves #124

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/packages/vsce/src/commands/enableCommands/enableTransactionCommand.ts
+++ b/packages/vsce/src/commands/enableCommands/enableTransactionCommand.ts
@@ -14,7 +14,6 @@ import { imperative } from "@zowe/zowe-explorer-api";
 import { commands, ProgressLocation, TreeView, window } from "vscode";
 import { CICSRegionTree } from "../../trees/CICSRegionTree";
 import { CICSTree } from "../../trees/CICSTree";
-import * as https from "https";
 import { CICSRegionsContainer } from "../../trees/CICSRegionsContainer";
 import { findSelectedNodes } from "../../utils/commandUtils";
 import { CICSTransactionTreeItem } from "../../trees/treeItems/CICSTransactionTreeItem";
@@ -46,7 +45,6 @@ export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<a
           });
           const currentNode = allSelectedNodes[parseInt(index)];
 
-          https.globalAgent.options.rejectUnauthorized = currentNode.parentRegion.parentSession.session.ISession.rejectUnauthorized;
 
           try {
             await enableTransaction(currentNode.parentRegion.parentSession.session, {
@@ -54,12 +52,10 @@ export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<a
               regionName: currentNode.parentRegion.label,
               cicsPlex: currentNode.parentRegion.parentPlex ? currentNode.parentRegion.parentPlex.getPlexName() : undefined,
             });
-            https.globalAgent.options.rejectUnauthorized = undefined;
             if (!parentRegions.includes(currentNode.parentRegion)) {
               parentRegions.push(currentNode.parentRegion);
             }
           } catch (error) {
-            https.globalAgent.options.rejectUnauthorized = undefined;
             window.showErrorMessage(
               `Something went wrong when performing an ENABLE - ${JSON.stringify(error, Object.getOwnPropertyNames(error)).replace(
                 /(\\n\t|\\n|\\t)/gm,
@@ -99,7 +95,7 @@ export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<a
   });
 }
 
-async function enableTransaction(session: imperative.AbstractSession, parms: { name: string; regionName: string; cicsPlex: string }): Promise<ICMCIApiResponse> {
+async function enableTransaction(session: imperative.AbstractSession, parms: { name: string; regionName: string; cicsPlex: string; }): Promise<ICMCIApiResponse> {
   const requestBody: any = {
     request: {
       action: {
--- a/packages/vsce/src/commands/enableCommands/enableTransactionCommand.ts
+++ b/packages/vsce/src/commands/enableCommands/enableTransactionCommand.ts
@@ -18,6 +18,7 @@ import { CICSRegionsContainer } from "../../trees/CICSRegionsContainer";
 import { findSelectedNodes } from "../../utils/commandUtils";
 import { CICSTransactionTreeItem } from "../../trees/treeItems/CICSTransactionTreeItem";
 import { CICSCombinedTransactionsTree } from "../../trees/CICSCombinedTrees/CICSCombinedTransactionTree";
+import { ICommandParams } from "../ICommandParams";
 
 export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<any>) {
   return commands.registerCommand("cics-extension-for-zowe.enableTransaction", async (clickedNode) => {
@@ -95,7 +96,7 @@ export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<a
   });
 }
 
-async function enableTransaction(session: imperative.AbstractSession, parms: { name: string; regionName: string; cicsPlex: string; }): Promise<ICMCIApiResponse> {
+async function enableTransaction(session: imperative.AbstractSession, parms: ICommandParams): Promise<ICMCIApiResponse> {
   const requestBody: any = {
     request: {
       action: {
--- a/packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
+++ b/packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
@@ -20,7 +20,7 @@ import { CicsCmciConstants } from "@zowe/cics-for-zowe-sdk";
 import { toEscapedCriteriaString } from "../../utils/filterUtils";
 import { CICSRegionsContainer } from "../CICSRegionsContainer";
 import { TextTreeItem } from "../treeItems/utils/TextTreeItem";
-import { getIconOpen } from "../../utils/profileUtils";
+import { getIconOpen, getIconPathInResources } from "../../utils/profileUtils";
 import { imperative } from "@zowe/zowe-explorer-api";
 
 export class CICSCombinedProgramTree extends TreeItem {
--- a/packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
+++ b/packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
@@ -21,6 +21,7 @@ import { toEscapedCriteriaString } from "../../utils/filterUtils";
 import { CICSRegionsContainer } from "../CICSRegionsContainer";
 import { TextTreeItem } from "../treeItems/utils/TextTreeItem";
 import { getIconOpen } from "../../utils/profileUtils";
+import { imperative } from "@zowe/zowe-explorer-api";
 
 export class CICSCombinedProgramTree extends TreeItem {
   children: (CICSProgramTreeItem | ViewMore)[] | [TextTreeItem] | null;
@@ -49,9 +50,8 @@ export class CICSCombinedProgramTree extends TreeItem {
         cancellable: true,
       },
       async (_, token) => {
-        token.onCancellationRequested(() => {
-          console.log("Cancelling the load");
-        });
+        token.onCancellationRequested(() => { });
+        let recordsCount: number;
         try {
           let criteria;
           if (this.activeFilter) {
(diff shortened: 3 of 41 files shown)


packages/vsce/src/trees/treeItems/CICSLibraryTreeItem.ts
```diff
@@ -58,35 +58,28 @@ export class CICSLibraryTreeItem extends TreeItem {
...
-      https.globalAgent.options.rejectUnauthorized = undefined;
-      const datasetArray = Array.isArray(libraryResponse.response.records.cicslibrarydatasetname)
-        ? libraryResponse.response.records.cicslibrarydatasetname
-        : [libraryResponse.response.records.cicslibrarydatasetname];
-      this.label = `${this.library.name}${this.parentRegion.parentPlex ? ` (${this.library.eyu_cicsname})` : ""}${
-        this.activeFilter ? ` (${this.activeFilter}) ` : " "
-      }[${datasetArray.length}]`;
+      const datasetArray = toArray(libraryResponse.response.records.cicslibrarydatasetname);
+      this.label = `${this.library.name}${this.parentRegion.parentPlex ? ` (${this.library.eyu_cicsname})` : ""}${this.activeFilter ? ` (${this.activeFilter}) ` : " "
+        }[${datasetArray.length}]`;
```
davenice, line 70: this is a bit tricky and the whitespace has changed with no real benefit to the reader.

what about something like this? i guess the linter still won't like it, but...

```suggestion
      const datasetArray = toArray(libraryResponse.response.records.cicslibrarydatasetname);
      this.label = `${this.library.name}
        ${this.parentRegion.parentPlex ? ` (${this.library.eyu_cicsname})` : ""}
        ${this.activeFilter ? ` (${this.activeFilter}) ` : " "}
        [${datasetArray.length}]`;
```

davenice: There are similar examples elsewhere, we could choose a common approach (whatever it is).

davenice (commented):
Couple of thoughts - look generally good though.

packages/vsce/src/trees/treeItems/CICSLibraryTreeItem.ts
AndrewTwydell: Had a go at resolving in [this commit](https://github.com/zowe/cics-for-zowe-client/pull/173/commits/37980ec8493398fbc434908bc1b893630905bc41) - what do you think?

davenice: I think that is way easier to maintain - guarantee it would take several goes to get the syntax, brackets, and backticks lined up any time I changed the previously implementation.

zFernand0 (approved):
Overall, LGTM! 😋 

Thanks for fixing up all the typing issues. 🙏 
And also for creating the common toArray function 🥳 
Finally, thanks a lot for removing `axios`! 🙏 

I know the primary purpose was to avoid messing with the reject unauthorized configuration of the https package, but I'm really happy with the little refactoring that happened here 🙏

zFernand0:
> `@zFernand0` I'd like your opinion on this as it resolves an issue you found relating to self-signed certs.

My apologies for taking so long, 😅 
But I really like what's going on in this PR 🙏 

> I intend to squash my messy commits before merging 👍🏼

Similar to Zowe Explorer, I think we don't have to worry about constantly force-pushing to your own branch and instead, we can choose to squash PRs when merging (as Dave suggested on another PR conversation/thread).

Great work on this! 🙏

zFernand0:
> With the current lack of automated testing (WIP), I'm recruiting as many people as possible to manually test this (`@davenice, @enamkhan, @chinmdas`, amongst others) but the more the merrier!

I'd love to see other people's thoughts/comments about their review/testing as well 🙏

packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
```diff
@@ -49,9 +50,8 @@ export class CICSCombinedProgramTree extends TreeItem {
         cancellable: true,
       },
       async (_, token) => {
-        token.onCancellationRequested(() => {
-          console.log("Cancelling the load");
-        });
+        token.onCancellationRequested(() => { });
```
enamkhan, line 53: Maybe () => true is fractionally neater.

enamkhan: Otherwise LGTM!

.editorconfig
```diff
@@ -3,7 +3,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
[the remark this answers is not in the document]
AndrewTwydell: From our experience (sometimes painful when they auto format on save back to indent of 4!) files already adhere. However that does not mean we shouldn't configure prettier on a format command or something 👍 

The reason for this change was to avoid the auto formatting moving the whole file around which will work in the meantime before we get a format command in. 🙂

AndrewTwydell:
> Similar to Zowe Explorer, I think we don't have to worry about constantly force-pushing to your own branch and instead, we can choose to squash PRs when merging (as Dave suggested on another PR conversation/thread).
> 
I'm a fan of a tidy commit history so a habit I've found myself in! More for me than anything else 😂

packages/vsce/src/commands/enableCommands/enableTransactionCommand.ts
```diff
@@ -99,7 +95,7 @@ export function getEnableTransactionCommand(tree: CICSTree, treeview: TreeView<a
   });
 }
 
-async function enableTransaction(session: imperative.AbstractSession, parms: { name: string; regionName: string; cicsPlex: string }): Promise<ICMCIApiResponse> {
+async function enableTransaction(session: imperative.AbstractSession, parms: { name: string; regionName: string; cicsPlex: string; }): Promise<ICMCIApiResponse> {
```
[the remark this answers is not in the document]
AndrewTwydell: > Curious if we can type it as one
>
Good plan, I'll add that onto this PR while we're already touching those methods.

> Also, I know I have a draft PR (#14) about migrating a lot of these SDK-like functions to the `packages/sdk` but I'm afraid that such PR might be a bit ambitious and also out of date 😋
> 
Now I'm more familiar with the monorepo I'll look over what you've got and comment over there. Seems silly to waste effort you've already done! 😎

davenice:
I've had a go with:
- two separate hosts in play for z/OSMF, one with proper cert and one self-signed - both secured
- cicsplexes on one of the hosts - one with security (incl self signed ssl) and one without security

I can access all of these systems at once without an issue.

i'm seeing a possible regression around what happens if you connect to a plex but specify a regionName - it seems to be displaying the "All Programs" and similar menus, and nesting everything as if it's a whole plex.

image

packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedProgramTree.ts
AndrewTwydell: We can do this in a follow-up refactor if that's okay! 😄

.editorconfig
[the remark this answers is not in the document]
davenice: Yes - we're having quite a few issues with whole files changing, and having to ask people to undo some of the changes they didn't intend (also issues with the changelog causing a conflict with every PR!)

If preferred, we could set this to 2 in a separate PR and format EVERYTHING so that we get that out of the way in one bash.

AndrewTwydell:
Good find @davenice! Will take a look now... Is that where you've specified a plex and region in the profile, not through any filtering

davenice:
> Good find @davenice! Will take a look now... Is that where you've specified a plex and region in the profile, not through any filtering

yes that's right - I think the issue is in profileManagement where we're deciding whether we've specified a region or group, the code is thinking we specified a group.

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -52,273 +53,292 @@ export class ProfileManagement {
...
+      });
+    } catch (error) {
+      if (error instanceof imperative.ImperativeError) {
+        if (!error.mDetails.msg.toUpperCase().includes("NODATA")) {
+          throw error;
+        }
+      }
+    }
+
+    return checkIfSystemGroup?.response.resultsummary.recordcount !== "0";
```
davenice, line 90: If we get an error (like NODATA), checkIfSystemGroup is still undefined, which means it isn't "0", which means we return true ... which is wrong.

I think the best thing to do is assign the recordcount to a variable and make sure it's defined. If it is, and it's non null - we got ourselves a result. Another option would be to parse it as an int and check it was greater than zero, but I suspect this will have its own issues.

```suggestion
    var recordCount = checkIfSystemGroup?.response.resultsummary.recordcount;
    return typeof recordCount !== "undefined" && recordCount !== "0";
```

.editorconfig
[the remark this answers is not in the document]
zFernand0: I like the idea of doing it on a separate PR 🙏 
Thank you both for considering this! 🙏

packages/vsce/src/utils/profileManagement.ts
davenice: After a discussion with Andrew we came up with a clearer option... I'll suggest that separately.

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -52,273 +53,292 @@ export class ProfileManagement {
...
+  }
```
davenice, line 91: If we get an error (like NODATA), checkIfSystemGroup is still undefined, which means it isn't "0", which means we return true ... which is wrong.

Making the flow clearer and ensuring we set group to false if checkIfSystemGroup is undefined will give the right result.

```suggestion
  public static async regionIsGroup(session: Session, profile: imperative.IProfile): Promise<boolean> {
    let isGroup = false;

    try {
      const checkIfSystemGroup: ICMCIApiResponse = await getResource(session, {
        name: "CICSRegionGroup",
        cicsPlex: profile.cicsPlex,
        regionName: profile.regionName,
        criteria: `GROUP=${profile.regionName}`,
      });

      if (checkIfSystemGroup && checkIfSystemGroup.response.resultsummary.recordcount !== "0") {
        isGroup = true;
      }
    } catch (error) {
      if (error instanceof imperative.ImperativeError) {
        if (!error.mDetails.msg.toUpperCase().includes("NODATA")) {
          throw error;
        }
      }
    }

    return isGroup;
  }
```

davenice:
OK - so I think we've resolved the issues we know about. I've tested with SMSS now (had forgotten that before).

Is there much else remaining to snag on this, @AndrewTwydell? I aspire to us getting this in this week ... if it's safe :-)

AndrewTwydell:
Nothing more that's been mentioned or found, and lots of manual testing has now been done. Seems I've got some fun conflicts to resolve later and we can hopefully get it in this afternoon after a final nod from the reviewers 🙂 Be very good to finally have this problem sorted!!!

zFernand0:
> We're pretty confident that this is an improvement now (within the bounds that there isn't much automated testing).
> 
> We'd like to get this in - are you happy?

**_100%_** 
Thank you all for giving this a try and special shoutout to Andrew for the refactoring (and removing `axios` 🙏  )

Issue #167, opened 2024-12-13
Should not be able to query SIT parameters on an Inactive ManagedRegion


AndrewTwydell:
**Describe the bug**

When connected to a plex with active and inactive managed regions, they show up with different icons. Inactive regions still have the ability to right click, Show SIT parameters. We should add a _when_ clause to that menu item to only show when regions are active, and add a check to the CMCI call to prevent the error.

**Additional context**

<img width="441" alt="Screenshot 2024-12-13 at 4 31 02 PM" src="https://github.com/user-attachments/assets/00ddd40a-c484-4ed7-b3aa-6796d532b7b5" />

davenice:
I had a little look at this, @enamkhan - didn't quite get as far as deciding how I'd want to manage adding extra bits of context string on, though so I hadn't !

https://github.com/davenice/cics-for-zowe-client/tree/davenice-only-show-sitparms-on-active-region

I think this change just hides the option on regions that are not active. I didn't prevent CMCI from making the request.

PR #183, opened 2024-12-20
Remove the sit param menu option when a region is inactive

enamkhan:
**What It Does**
Currently the "Show SIT Parameters" menu option can be selected on inactive regions which will cause an error. I've changed this so that this option is greyed out when the region is inactive. I've also added a check in the command so that the error does not surface when executed from the command palette.

**How to Test**
Connected to a plex with both active and inactive regions and verified that the menu item for 'Show SIT Parameters' is grayed out when the region is inactive. This will also be the case if multi-select is used and includes an inactive region.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Resolves https://github.com/zowe/cics-for-zowe-client/issues/167

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
 
 ## Recent Changes
-- BugFix: Now the 'Show SIT Parameters' menu option is grayed out for inactive regions. [#167](https://github.com/zowe/cics-for-zowe-client/issues/167)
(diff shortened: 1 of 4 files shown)


enamkhan:
Was using enablement to grey out the SIT Param menu option, however this had a side-effect of not showing the option in the command palette. So have changed to just not showing the menu option when the region is inactive.

AndrewTwydell (approved):
Looks good - thanks for fixing the bug! 😃

zFernand0:
Combining with 
- https://github.com/zowe/cics-for-zowe-client/pull/192

PR #189, opened 2025-01-03
Now URI encoding the region and cicsplex names

### LINKED ISSUE #178: Unable to get a list of CICS objects for a CICSPlex with "#" character in the end of its name
There are three CICSPlex CMCI using the same host name and tcp port.
Neither "cicsPlex" nor "regionName" is defined in a zowe team config.
The IBM CICS for Zowe Explorer extension connects to the host and all three CICSPlex names appear in the VSCode side bar "CICS" under "cics" profile.
But the only two with regular English characters work well. The one "CPLEXH#", that has "#" character in the end of its name, does not work.
Trying to uncollapse it in the side bar the extension does not give the list of CICS objects.

All three CICSPlex are accessible from IBM Explorer for z/OS with no issues.
Could it be that "#" caused the problem?
Please check. Let us know if I need to ask the client for any diagnostic information.

IBM CICS for Zowe Explorer extension version is 3.0.1

enamkhan:
**What It Does**
Issues was caused by not encoding the # so have updated the getResourceUri method so that both the region and cicsplex names are now uri encoded. 

**How to Test**
Created a region with a # in the name and connected with updated extension. Have also added some unit tests to cover this case. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Resolves #178


davenice:
Thanks Enam - one thing worth mentioning is that I have seen a customer using system groups with hashes in - system groups don't support wildcards so they made their own which they maintained manually 😬 

I don't think there is a lot of support for system groups in the GUI - the main one I've seen is that if you specify a region in your config, you can specify a system group instead. I think this will work with your changes because I think the code just sees it as a region name rather than a region group - do you think we've covered all the places this might need addressing?

packages/sdk/src/utils/Utils.ts
```diff
@@ -31,34 +31,36 @@ export class Utils {
...
 
-    if (options && options.criteria) {
-      cmciResource += `${delimiter}${CicsCmciConstants.CRITERIA}=${this.enforceParentheses(encodeURIComponent(options.criteria))}`;
-      delimiter = "&";
-    }
+    if (options) {
+      if (options.criteria) {
+        cmciResource += `${delimiter}${CicsCmciConstants.CRITERIA}=${this.enforceParentheses(encodeURIComponent(options.criteria))}`;
+        delimiter = "&";
+      }
```
AndrewTwydell, line 43: Now could be a good time to implement Fernando's efficiency syntax! 💪🏼 

```suggestion
if (options?.criteria) {
    cmciResource += `${delimiter}${CicsCmciConstants.CRITERIA}=${this.enforceParentheses(encodeURIComponent(options.criteria))}`;
    delimiter = "&";
}
// Same for all others
```

packages/sdk/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the IBM® CICS® Plug-in for Zowe CLI will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Now URI encoding the region and cicsplex names. [#178](https://github.com/zowe/cics-for-zowe-client/issues/178)
```
AndrewTwydell, line 7:
```suggestion
- BugFix: URI encoding the region and cicsplex names. [#178](https://github.com/zowe/cics-for-zowe-client/issues/178)
```

packages/sdk/src/utils/Utils.ts
```diff
@@ -31,34 +31,36 @@ export class Utils {
 
     let delimiter = "?"; // initial delimiter
 
-    const cicsPlex = (options && options.cicsPlex) == null ? "" : `${options.cicsPlex}/`;
-    const region = (options && options.regionName) == null ? "" : options.regionName;
+    const cicsPlex = (options && options.cicsPlex) == null ? "" : `${encodeURIComponent(options.cicsPlex)}/`;
+    const region = (options && options.regionName) == null ? "" : encodeURIComponent(options.regionName);
```
AndrewTwydell, line 35: Are these `== null` for a reason, or can we just do `options?.cicsPlex ? "" : ...`??

```suggestion
    const cicsPlex = options?.cicsPlex ? "" : `${encodeURIComponent(options.cicsPlex)}/`;
    const region = options?.regionName ? "" : encodeURIComponent(options.regionName);
```

packages/vsce/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Updated commands to us Utils.getResourceURI. [#178](https://github.com/zowe/cics-for-zowe-client/issues/178)
```
```suggestion
- BugFix: Updated commands to use Utils.getResourceURI. [#178](https://github.com/zowe/cics-for-zowe-client/issues/178)
```

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -104,4 +104,8 @@ export const CicsCmciConstants = {
      */
   CICS_CMCI_EXTERNAL_RESOURCES: ["CICSLocalTransaction", "CICSRemoteTransaction", "CICSDefinitionTransaction", "CICSLocalFile"],
```
AndrewTwydell, line 105:
```suggestion
CICS_CMCI_TRANSACTION_DEFINITION: "CICSDefinitionTransaction",
CICS_CMCI_LOCAL_TRANSACTION: "CICSLocalTransaction",
CICS_CMCI_REMOTE_TRANSACTION: "CICSRemoteTransaction",
CICS_CMCI_LOCAL_FILE: "CICSLocalFile",

CICS_CMCI_EXTERNAL_RESOURCES: [
  CicsCmciConstants.CICS_CMCI_LOCAL_TRANSACTION,
  CicsCmciConstants.CICS_CMCI_REMOTE_TRANSACTION,
  CicsCmciConstants.CICS_CMCI_TRANSACTION_DEFINITION,
  CicsCmciConstants.CICS_CMCI_LOCAL_FILE
],
```

AndrewTwydell: Mentioned below, seems an odd way to structure constants.

packages/vsce/src/commands/closeLocalFileCommand.ts
```diff
@@ -143,18 +143,13 @@ async function closeLocalFile(
...
-    "?CRITERIA=(FILE=" +
-    parms.name +
-    ")";
+  const options: IGetResourceUriOptions = {
+    "cicsPlex": parms.cicsPlex,
+    "regionName": parms.regionName,
+    "criteria": `FILE='${parms.name}'`
+  };
+
+  const cmciResource = Utils.getResourceUri(CicsCmciConstants.CICS_CMCI_EXTERNAL_RESOURCES[3], options);
```
AndrewTwydell, line 152: I know this was not this PR's change but why are we listing the resources in an array of strings? This does not seem overly helpful when referring to constrants 🤔 

Have tried an example of how we could structure it to make it more _constant_.

AndrewTwydell: That way, we could use `CicsCmciConstants.CICS_CMCI_LOCAL_FILE`

AndrewTwydell (changes requested):
Looks great - few things to look at 😄

packages/sdk/src/constants/CicsCmci.constants.ts
AndrewTwydell: Not changed in this PR but noticed because you're using the array of strings

packages/vsce/src/commands/closeLocalFileCommand.ts
enamkhan: I'm not sure why this was the case.

packages/sdk/src/utils/Utils.ts
enamkhan: Will try, bit the reason I changed this was to get around a lint complexity error.

AndrewTwydell (approved):
LGTM 👍🏼

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -100,8 +100,27 @@ export const CicsCmciConstants = {
   PARAMETER: "PARAMETER",
 
   /**
-     * The CICS CMCI external resource names
+     * The CICS CMCI transaction definition
      */
-  CICS_CMCI_EXTERNAL_RESOURCES: ["CICSLocalTransaction", "CICSRemoteTransaction", "CICSDefinitionTransaction", "CICSLocalFile"],
+  CICS_CMCI_TRANSACTION_DEFINITION: "CICSDefinitionTransaction",
```
zFernand0, line 105: For backward compatibility reasons, could we keep the `CICS_CMCI_EXTERNAL_RESOURCES`.
And add a `@deprecated` line in the JSDoc 🙏 

```
/**
     * The CICS CMCI external resource names
     * @deprecated
     */
     CICS_CMCI_EXTERNAL_RESOURCES: ["CICSLocalTransaction", "CICSRemoteTransaction", "CICSDefinitionTransaction", "CICSLocalFile"],
```

zFernand0 (changes requested):
Overall, this LGTM! 😋 

I do have one small request to keep the old CMCI.constant to avoid possible breaking changes 🙏

PR #193, opened 2025-01-06
update regions icon when auto-expanded

### LINKED ISSUE #194: Regions icon not updating to open when plex expanded
When we expand a Plex, we auto-expand the Regions tree. This expansion does NOT update the icon to be an open folder. However, when we close and manually expand the regions tree, the icon is updated.

https://github.com/user-attachments/assets/a5107e44-22b5-494e-8861-277dc3576186

AndrewTwydell:
**What It Does**
Resolves #194 
Updates the Regions container tree icon when auto-expanded by the Plex tree expanding.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -1,14 +1,15 @@
 # Change Log
 
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
-## `3.2.3`
-
-- Changed icons for: folder, local file, local transaction, program and task. Updated 16 icons for different status and themes. [#182](https://github.com/zowe/cics-for-zowe-client/pull/182)
-
 
 ## Recent Changes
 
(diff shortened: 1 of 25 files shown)


zFernand0 (approved):
Changes LGTM! 😋 

As a general question, should an empty folder still show as open when expanded?
Also, should we move the `iconOpen` function to a common location since other container-like classes might make use of something similar as well.

davenice:
> As a general question, should an empty folder still show as open when expanded? 

I'm thinking .... yes? As a user I want to know that the folder is open and empty. Otherwise I'm liable to think the developers made a bug and it just .... won't ..... open? :-) 🤔

davenice:
Looks good! I'll merge (I don't think we need to ship this today because we have a few other fixes on the way imminently)

PR #201, opened 2025-01-09
Catch NODATA response when generating cache

### LINKED ISSUE #200: Filtering All Resources trees errors if none are found
When searching in the _All Resources_ trees, if a pattern does not match any resources, a NODATA error is sent to the user. This is a regression as it used to present a message stating no resources were found.

Screenshot 2025-01-09 at 10 54 55 PM

AndrewTwydell:
**What It Does**
Resolves #200 

Adds a check for a NODATA response when generating a cache token for the combined trees. If the query comes back with NODATA, an error is thrown by the SDK. This was not handled and ignored by the VSCE.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

Screenshot 2025-01-09 at 10 59 38 PM

### DIFF
--- a/packages/vsce/src/utils/profileManagement.ts
+++ b/packages/vsce/src/utils/profileManagement.ts
@@ -312,21 +312,29 @@ export class ProfileManagement {
     group?: string
   ): Promise<{ cacheToken: string; recordCount: number; }> {
     const session = this.getSessionFromProfile(profile.profile);
-    const allProgramsResponse = await getResource(session, {
-      name: resourceName,
-      cicsPlex: plexName,
-      ...group ? { regionName: group } : {},
-      criteria: criteria,
-      queryParams: {
-        summonly: true,
-        nodiscard: true,
-        overrideWarningCount: true,
+    try {
+      const allProgramsResponse = await getResource(session, {
+        name: resourceName,
+        cicsPlex: plexName,
+        ...group ? { regionName: group } : {},
+        criteria: criteria,
+        queryParams: {
+          summonly: true,
+          nodiscard: true,
+          overrideWarningCount: true,
+        }
+      });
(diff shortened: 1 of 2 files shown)


davenice (approved):
I got distracted while looking at this.

It looks good to me. Simpler than our late night marathon last week because most of the logic is similar, we just need to catch the error! And we've already got code to return a blank.

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -312,21 +312,29 @@ export class ProfileManagement {
...
+        }
       }
-    });
-    if (allProgramsResponse.response.resultsummary.api_response1_alt === "OK") {
-      if (allProgramsResponse.response && allProgramsResponse.response.resultsummary) {
-        const resultsSummary = allProgramsResponse.response.resultsummary;
-        return { cacheToken: resultsSummary.cachetoken, recordCount: parseInt(resultsSummary.recordcount, 10) };
+    } catch (error) {
+      if (error instanceof imperative.ImperativeError) {
+        if (!error.mDetails.msg.toUpperCase().includes("NODATA")) {
```
zFernand0, line 335: I like the simple implementation.
I'm curious if we can look for something a bit more deterministic than just `NODATA`
Is there a specific HTTP code, or errorCode coming from the CMCI REST API?

zFernand0 (approved):
LGTM! 😋

packages/vsce/src/utils/profileManagement.ts
zFernand0: Disregard 😅 

---

After reading a bit more doc, NODATA seems to be standard for the CMCI APIs 😅

davenice: Sad but true! 👍👍👎👎

PR #202, opened 2025-01-09
Refactor command registration and tree expansion events

AndrewTwydell:
**What It Does**
Refactor `extension.ts` to reduce repetition. The majority of the expansion events were 1 of 2 methods, with a couple of exceptions. The refactor ensures the common behaviour share a single method.

I also moved the commands registration out of the file to it's own method for tidiness.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

### DIFF
--- a/packages/vsce/src/commands/disableCommands/index.ts
+++ b/packages/vsce/src/commands/disableCommands/index.ts
@@ -1,0 +1,15 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


davenice (approved):
I think this looks sensible. I like beginning to split out extension.ts because it will reduce the number of merge conflicts we hit.

I like the map of types to actions.

I've tested this locally, expanding a full tree in a CICSplex and in an SMSS region.

zFernand0 (approved):
I really like how clean the `extension.ts` file looks after refactoring 🙏 

LGTM! 😋

PR #204, opened 2025-01-10
Using real data for unit test mocking

AndrewTwydell:
**What It Does**
Adapts SDK unit tests to use real CMCI responses for a more accurate behaviour.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/sdk/__tests__/__mocks__/CmciGetResponse.ts
+++ b/packages/sdk/__tests__/__mocks__/CmciGetResponse.ts
@@ -1,0 +1,200 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { readFileSync } from "fs";
+import { join } from "path";
+
+export const nodataXmlResponse = readFileSync(join(__dirname, "nodata.xml")).toString();
+export const ok1RecordXmlResponse = readFileSync(join(__dirname, "ok.1_record.xml")).toString();
+export const ok2RecordsXmlResponse = readFileSync(join(__dirname, "ok.2_records.xml")).toString();
+export const okCacheXmlResponse = readFileSync(join(__dirname, "ok.cache.xml")).toString();
+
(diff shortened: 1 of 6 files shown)


davenice (approved):
This looks good!

Great to use actual data.

We might find in the future we need to have specific data per test, rather than a generic set of mocks, but I imagine that the generic set will do us for a while!

AndrewTwydell:
@zFernand0 I'd like to run this by you, now that I've marked it ready for review - the only potential contentious point would be that I've changed the method we're mocking here.

We're unit testing `getResource` still, but I'm now mocking the XML response rather than the JSON parsed version of the XML, so if anything I'm now testing more?? The idea behind this was so we can test real CMCI error messages more easily, but I'd like to know if you're onboard with this style of testing or not! 😀

packages/sdk/__tests__/__unit__/get/Get.resource.unit.test.ts
```diff
@@ -90,13 +92,13 @@ describe("CMCI - Get resource", () => {
...
       expect(error).toBeDefined();
-      expect(error.message).toContain("Required parameter 'CICS Resource name' must not be blank");
+      expect(error?.message).toContain("Required parameter 'CICS Resource name' must not be blank");
     });
   });
 
   describe("success scenarios", () => {
 
-    const deleteSpy = jest.spyOn(CicsCmciRestClient, "getExpectParsedXml").mockResolvedValue(content);
+    const getExpectStringMock = jest.spyOn(CicsCmciRestClient, "getExpectString").mockResolvedValue(ok2RecordsXmlResponse);
```
almightyrush, line 101: No need to add mockResolvedValue() here. You have already added in beforeEach() method.
`const getExpectStringMock = jest.spyOn(CicsCmciRestClient, "getExpectString")`

AndrewTwydell: Great, thanks! All done 😄

traeok (approved):
LGTM, thanks @AndrewTwydell!

packages/sdk/__tests__/__mocks__/CmciGetResponse.ts
zFernand0, line 1: I really like this approach 🙏 
Thanks for adding the mock responses 🙏

zFernand0 (commented):
LGTM! 😋 
---

Apologies for the late review 😓

PR #205, opened 2025-01-10
SDK getResource failOnNoData flag

AndrewTwydell:
**What It Does**
Adds a failOnNodata flag, defaulting to true (which is the current behaviour), to dictate if CMCI sending a NODATA response should throw an error or not.

* PR relies on #204 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/utils/profileManagement.ts
+++ b/packages/vsce/src/utils/profileManagement.ts
@@ -9,10 +9,15 @@
  *
  */
 
-import { getCache, getResource } from "@zowe/cics-for-zowe-sdk";
+import {
+  CicsCmciConstants,
+  CicsCmciRestError,
+  getCache,
+  getResource,
+  IResourceQueryParams
+} from "@zowe/cics-for-zowe-sdk";
 import { Session } from "@zowe/imperative";
-import { imperative, Types, ZoweVsCodeExtension } from "@zowe/zowe-explorer-api";
-import { window } from "vscode";
+import { Gui, imperative, MessageSeverity, Types, ZoweVsCodeExtension } from "@zowe/zowe-explorer-api";
 import { CICSPlexTree } from "../trees/CICSPlexTree";
 import { toArray } from "./commandUtils";
 import constants from "./constants";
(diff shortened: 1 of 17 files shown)


davenice:
This looks good too - do we need it for use in VSCE somewhere?

failOnNoData is added, but is optional, right? Is there any way this could break existing callers?

davenice:
The code looks good as far as I can see. I think we normally call it RESP/RESP2 rather than RESPONSE_1 and RESPONSE_2 - so we'd talk about "your RESP code" or "your RESP2 code" - but I don't think that's a big problem and it does disambiguate a bit!

I appreciate having a unit test that's identical other than asking for the old and new style of error response. That's useful.  Are there any other kinds of errors we ought to drive that way? Is there anything other than "NODATA" we should prod?

I will sync the code and test it - but because this is and SDK change it would be valuable to have @zFernand0's input, or somebody that he delegates to!

zFernand0 (approved):
LGTM! 😋 
---

Thanks for adding the cmci request options! 🙏 
Seem like the failOnNoData is properly defaulted to `true` as not to break existing behavior on the CLI 🙏

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -165,20 +212,41 @@ export class ProfileManagement {
...
+      }
+    } catch (error) {
+      let errorMessage = `Error requesting CICSManagedRegion - ${JSON.stringify(error)}`;
+      if (
+        error instanceof CicsCmciRestError && (
+          error.RESPONSE_1 === CicsCmciConstants.RESPONSE_1_CODES.INVALIDPARM ||
+          error.RESPONSE_1 === CicsCmciConstants.RESPONSE_1_CODES.INVALIDDATA)) {
+        errorMessage = `Plex ${profile.cicsPlex} not found.`;
+      } else if (error instanceof CicsCmciRestError) {
+        errorMessage = `${error.RESPONSE_1_ALT} ${error.RESPONSE_2_ALT} requesting CICSManangedRegions.`;
```
davenice, line 242:
```suggestion
        errorMessage = `${error.RESPONSE_1_ALT} ${error.RESPONSE_2_ALT} requesting CICSManagedRegions.`;
```

davenice (approved):
Two minor wording tweaks and it's a yes from me

PR #207, opened 2025-01-13
Updating icons for disabled and closed statuses resources

Rajat-Sahrawat:
**What It Does**
Adds visual decoration for closed and disabled statuses for local files, local transactions and programs. Also updated visual consistency for icons for task states- running, suspended and dispatched.

**How to Test**
change status for the mentioned resources to see the updated icons

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/vsce/src/utils/profileUtils.ts
```diff
@@ -55,3 +55,30 @@ export function getIconOpen(open: boolean = true) {
     `folder-${open ? "open" : "closed"}-dark.svg`,
     `folder-${open ? "open" : "closed"}-light.svg`);
 }
+
+export function getIconByStatus(resourceType: string, resourceTreeItem: any) {
```
AndrewTwydell, line 59: A thought - and I know I've added icon-related methods to this file recently - but should we consider creating a new icons file under utils for this as well as the other icon methods (`getIconOpen`, `getIconPathInResources`, now `getIconByStatus`)?? I don't think they relate to profiles, and now we have more methods, perhaps you could make it's own file?? 

I'm thinking `vsce/src/utils/iconUtils.ts` or something similar. What do you think @almightyrush ??

davenice:
A few screenshots to give an idea. Visually looking great to me.

image

image

image

davenice:
Hey @almightyrush - please rebase on main and resolve the conflict with extension.ts 👍

davenice (commented):
I like the icons and the code looks good.

The test coverage less so...

A couple of things I've been thinking:
1) Every icon ends .dark.svg or .light.svg so I'm wondering if we pass the 'root' of the icon name in and push that part down lower ... would that make things better? I think it might.
2) An option for unit testing is we split into a method that generates the filename which we can easily verify with text comparison, and then we could check the file exists on disk. But maybe it's just as easy to have it return a file and verify it matches the file instance we were expecting.

What do you think, @almightyrush ? A bunch of this is refactor but we have added some new function so....

davenice (approved):
This looks good!

Thoughts for future PRs in this area...
1) A test that goes through each option and verifies that the right files exist on disk
2) I think we can simplify the logic a little bit now for local file in particular where we append `-disabled` or `-closed` rather than having all the nested ternary ifs we could just say "filename is local-file ... if we're disabled append -disabled, if we're closed append -closed". That's not a must-do for this PR in my eyes - just an idea that we could work on that logic in future.

davenice (changes requested):
On testing, I'm seeing the light image being used in the dark theme and vice versa - would be good to work out why. Sorry...!

almightyrush:
> On testing, I'm seeing the light image being used in the dark theme and vice versa - would be good to work out why. Sorry...!

Thanks for finding it out @davenice 👌🏽
It was a small confusion and it got missed by me while refactoring😛. 

`light: join(__dirname, "..", "resources", "imgs", iconFileName + "-dark.svg"),`
`dark: join(__dirname, "..", "resources", "imgs", iconFileName + "-light.svg"),`

Here the `dark.svg` means the icon is of dark colour which will be used for light theme. 
And I got confused here thinking the icons are `dark.svg` used for dark theme and vice versa.
Reverted the below changes with the above one.

`light: join(__dirname, "..", "resources", "imgs", iconFileName + "-light.svg"),`
`dark: join(__dirname, "..", "resources", "imgs", iconFileName + "-dark.svg"),`

packages/vsce/src/utils/iconUtils.ts
AndrewTwydell, line 1: Up for a discussion, but I'm thinking a couple of these method names are difficult to understand their exact purpose. What about something like...
- `getIconPathInResources` -> `getIconFilePathFromName(iconFileName)`
- `getIconOpen` -> `getFolderIcon(folderIsOpen)`

Don't think it makes a huge difference but maybe something easier to differentiate between each of the methods without having to look at their source 😄  Let me know what you think

packages/vsce/src/trees/CICSRegionTree.ts
AndrewTwydell, line 1: We could probably treat regions with the same new status logic as the resources we've changed right?? We're changing the icon based on `region.cicsstate` being `ACTIVE` or not, so maybe we could build that into the `getIconByStatus` method too?

packages/vsce/__tests__/__unit__/utils/iconUtils.unit.test.ts
```diff
@@ -0,0 +1,108 @@
...
+const iconPath = {
+  light: "program-dark.svg",
+  dark: "program-light.svg",
+};
+
+describe("Test suite for iconUtils", () => {
+  describe("Test suite for getIconPathResources", () => {
+    it("Should return icon path for dark and light theme", () => {
+      const result = getIconPathInResources("program");
+      expect(result.dark).toContain(iconPath.dark);
```
AndrewTwydell, line 12: Is `expect().toEqual()` more accurate than `.toContain()`?? We don't want it to contain anything other than that string right?

AndrewTwydell (commented):
Couple of notes / questions left of the PR, and I will manually test with all status possibilities with all resource types too 👍🏼

packages/vsce/__tests__/__unit__/utils/iconUtils.unit.test.ts
davenice: I hit this difficulty - it's quite hard to form the exact string for the path because `__dirName__` in the test folders is different from `__dirName__` in the production code folder. So I ended up doing what Rushabh is - I did look at using a match method to say "must end with this string" but again I didn't quite get that working neatly.

Another option would be to refactor again so that only the part after `__dirName__` was generated but I don't like this so much.

I think a future step to verify that each generated name had a matching image would help make things watertight, r too.

AndrewTwydell: Okay cool thanks for clarifying that - makes sense. Does a `toEndWith` method exist? Although it's barely better than contains so not a huge deal.

AndrewTwydell: I'm probably nit-picking, let's all just be happy we're adding test coverage! 😎

davenice (approved):

davenice:
My concerns are addressed - i'll just cancel my approval until those couple of small bits Andrew mentioned are decided on one way or the other.

davenice:
In fact I'll leave my review and we'll just wait for Andrew to had his. None of those bits are absolute dealbreakers, just improvements!

packages/vsce/src/utils/iconUtils.ts
almightyrush: Yes this sounds good.!!.

AndrewTwydell (approved):
Looks great, works great, nice improvements 😎

davenice (approved):

AndrewTwydell (approved):
Looks good, I've tested all the updated icons locally in light and dark theme, both look great!

zFernand0:
> `@zFernand0` - would be good to have a nod from you or someone on the ZE side!

Sorry for the delay on this 🙏 
I love the new icons 🙏

PR #213, opened 2025-01-23
show library datasets under correct library name

### LINKED ISSUE #212: Library Tree shows all library datasets in all library tree items
**Describe the bug**

When expanding a library resource, all library datasets (from all libraries) are listed, not just those in that library.

Should only list datasets in that library

Image

AndrewTwydell:
**What It Does**
Resolves #212 .

This changes the LibraryTreeItems to use a default filter of their Library name so only the Library datasets in that Library show up in that tree. Then it appends the active filter of the library to that filter to search the DSName attribute. 

Resulting behaviour is only the library datasets with the correct library name show up under that tree, with the filter applying to the dsname attribute.

**How to Test**

Have multiple Library resources with different datasets in. Expand each library. Before this update, they would show the same datasets - now they should show correct datasets. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

Updated view with bug fixed.
Screenshot 2025-01-27 at 10 44 31 AM

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 ## Recent Changes
 
 - Updated status icons for local transactions, local files, tasks and programs. [#203](https://github.com/zowe/cics-for-zowe-client/issues/203)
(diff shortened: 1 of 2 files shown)


almightyrush (approved):
LGTM! 🙏

davenice (approved):
Tested locally - looks good.

PR #216, opened 2025-01-28
WIP: Add support for using the ltpa token

enamkhan:
**What It Does**

https://github.com/zowe/cics-for-zowe-client/issues/217

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/extension.ts
+++ b/packages/vsce/src/extension.ts
@@ -121,6 +121,7 @@ export async function activate(context: ExtensionContext) {
       } catch (error) {
         const newSessionTree = new CICSSessionTree(
           node.element.getParent().profile,
+          node.element.getParent().getSession(),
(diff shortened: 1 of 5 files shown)


davenice (commented):
Thanks for getting started on this @enamkhan 👍 

I don't think we should add an option that talks about MFA. if we were going to add an option it might be around the functional desire "don't present my password a second time" or "don't store my password", but I think that might need more discussion with Zowe Explorer.

For this first part of support, let's look for an LTPA token coming back from the server, if there is one then cache it for the duration of the connection lifetime, and present it instead of basic auth.

I'd be keen to chat over the code with you! 👀

davenice:
It's also possible that there are difficulties here because we are using Zowe Explorer to make the connections, we don't have full control ourselves.... Perhaps you've mentioned that to me before 🤔 let's talk it over!

Issue #217, opened 2025-01-28
Use LTPA tokens to allow CMCI "sessions"


davenice:
The CMCI JVM server is available in most modern CICS installations - it puts a Liberty instance in front of the CMCI connection.

The was used by the Eclipse-based CICS Explorer as a route to provide MFA support, because the Liberty instance supports LTPA tokens out-of-the-box.

LTPA tokens give the concept of a session to the CMCI REST interface - after the initial authentication with basic auth, a cookie with an LTPA token is returned. If that LTPA token is presented (instead of basic auth) on subsequent REST calls, Liberty will make the request as the originally authenticated user.

This has a couple of key benefits...
1) Basic auth information is transferred across the wire less frequently, reducing the window for any attack to grab credentials
2) The connection becomes possible to use with MFA

This piece of work should cover:
- Connecting with basic auth
- In the response from the server, detecting an LTPA token has been returned
- Caching the LTPA token in memory
- Using the LTPA token on subsequent REST requests
- When a request using the LTPA token gets a 401 retrying with basic auth

davenice:
This is complete now - subject to autostore defects.

PR #219, opened 2025-01-29
Added support for ltpa token

enamkhan:
This pr is a work in progress to add support for a using the ltpa token.

https://github.com/zowe/cics-for-zowe-client/issues/217

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


enamkhan:
@davenice I'm just going to refactor now and remove the profile config.

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -282,9 +268,8 @@ export class ProfileManagement {
   /**
    * Return all the regions in a given plex
    */
-  public static async getRegionInfoInPlex(plex: CICSPlexTree): Promise<any[]> {
+  public static async getRegionInfoInPlex(plex: CICSPlexTree, session: Session): Promise<any[]> {
```
AndrewTwydell, line 271: We can probably get the session (2nd arg) from the plex tree (1st arg) using `getParent` or `getSession`, or a combination, right? Although it's not exactly messy currently so optional.

packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedWebServiceTree.ts
```diff
@@ -204,4 +209,8 @@ export class CICSCombinedWebServiceTree extends TreeItem {
   public getParent() {
     return this.parentPlex;
   }
+
+  public getSession() {
+    return this.getParent().getParent().getSession();
```
AndrewTwydell, line 214: Can this just be `this.getParent().getSession()` as the parent's `getSession` will also keep going up the tree??

AndrewTwydell: If so, same for all combination trees

AndrewTwydell (commented):
In a lot of the resource trees (CICSProgramTree), we're not using the new getters to get the session - might be nice if possible?

Might be nice to be able to do `this.getSession()` here like the combination trees.

https://github.com/zowe/cics-for-zowe-client/blob/df6656b9a0aa867542d22929fe43f533fd8c4344/packages/vsce/src/trees/CICSProgramTree.ts#L48

packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedTCPIPServiceTree.ts
zFernand0, line 1: It feels like we have to make the same/similar changes across all/most CICSCombinedTrees.
Wondering if there is some consolidation that can be done (in a future PR) to create a BaseCombinedTree class that the others extend :yum:

packages/vsce/src/trees/CICSSessionTree.ts
```diff
@@ -28,15 +29,19 @@ export class CICSSessionTree extends TreeItem {
     super(profile.name, TreeItemCollapsibleState.Collapsed);
     this.children = [];
     this.contextValue = `cicssession.${profile.name}`;
+
     this.session = new imperative.Session({
-      type: "basic",
+      type: SessConstants.AUTH_TYPE_TOKEN,
+      storeCookie: true,
```
zFernand0, line 35: I know I mentioned that perhaps we could move some of the LTPA token support to the CLI. 
However, it makes sense the way this is being done here since (as you mentioned on the last standup) this is all handled by Imperative.

Curious if the CICs plug-in could make use the the LTPA token when the CLI is in Daemon mode. 😋 
No need to worry about testing this, since people haven't asked for this feature yet 😋

zFernand0 (commented):
LGTM! 😋 
---

Haven't had a chance to test this, but the changes make sense to me! 😋

packages/vsce/src/trees/CICSCombinedTrees/CICSCombinedTCPIPServiceTree.ts
AndrewTwydell: This refactor is actually in progress in [this PR](https://github.com/zowe/cics-for-zowe-client/pull/214). Because all the resource and combined resource trees are practically identical, we can use common tree classes for them all :D

packages/vsce/src/utils/resourceUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+        ...profile?.cicsPlex && { cicsPlex: profile.cicsPlex },
+        ...regionName && { regionName: regionName},
+        ...cicsPlex && { cicsPlex: regionName },
+        ...params?.criteria && { criteria: params.criteria },
+        ...params?.parameter && { parameter: params.parameter },
+        ...params?.queryParams && { queryParams: params.queryParams },
+      },
+      { ...requestBody,
+        ...{ failOnNoData: false, useCICSCmciRestError: true }
+      });
```
AndrewTwydell, line 44: This seems complex. Why are we passing optional `profile` and `regionName`/`cicsPlex` args? And then set the regionName twice? Probably better passing in just the regionName and cicsPlex as optionals, and then give them directly to getResource (if you don't provide them, they will be undefined anyway)

packages/vsce/src/utils/resourceUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+import { getErrorCode } from "./errorUtils";
+
+export async function runGetResource({ session, resourceName, profile, regionName, cicsPlex, params }: {
+  session: Session,
+  resourceName: string,
+  profile?: imperative.IProfile,
+  regionName?: string,
+  cicsPlex?: string,
+  params?: { criteria?: string, parameter?: string; queryParams?: IResourceQueryParams; };
+}, requestBody?: {}) {
```
AndrewTwydell, line 28: This suggests `requestBody` has a type of empty object doesn't it? What can requestBody actually be?

packages/vsce/tsup.config.json
```diff
@@ -10,6 +10,5 @@
   "external": [
     "vscode"
   ],
-  "minify": "terser",
```
AndrewTwydell, line 13: Development change snuck in here!

packages/vsce/src/utils/resourceUtils.ts
enamkhan: This is used in the commands so for example can be,
{request: {
      action: {
        $: {
          name: "ENABLE",
        },
      },
    }
  }

packages/vsce/src/utils/resourceUtils.ts
enamkhan: I wanted to support both profile and just providing the attributes, maybe best to have two functions with one calling the other.

packages/vsce/CHANGELOG.md
```diff
@@ -4,9 +4,11 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 
 ## Recent Changes
 
+- Enhancement:Use LTPA tokens to allow CMCI "sessions" [#217](https://github.com/zowe/cics-for-zowe-client/issues/217)
```
```suggestion
- Enhancement: Use LTPA tokens to allow CMCI "sessions" [#217](https://github.com/zowe/cics-for-zowe-client/issues/217)
```

packages/sdk/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the IBM® CICS® Plug-in for Zowe CLI will be documented in this file.
 
+
+## Recent Changes
+
+- Enhancement: Use LTPA tokens to allow CMCI "sessions". [#217](https://github.com/zowe/cics-for-zowe-client/issues/217)
```
AndrewTwydell, line 8: I'm not sure adding some constants to the package warrants a changelog entry talking about LTPA tokens. I think we should change this to be something along the lines of...
```suggestion
- Enhancement: Added CICS resource names to available constants. [#217](https://github.com/zowe/cics-for-zowe-client/issues/217)
```

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -18,6 +18,11 @@ export const CicsCmciConstants = {
      */
   CICS_SYSTEM_MANAGEMENT: "CICSSystemManagement",
 
+  /**
+     * Specifies the required part of the REST interface URI to access initialization parameter
```
AndrewTwydell, line 22:
```suggestion
     * Specifies the required part of the REST interface URI to access system initialization parameters
```

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -48,6 +58,11 @@ export const CicsCmciConstants = {
      */
   CICS_DEFINITION_WEBSERVICE: "CICSDefinitionWebService",
 
+  /**
+     * Specifies the required part of the REST interface URI to access tcp/ip service definitions
+     */
+  CICS_DEFINITION_TCPIPSERVICE: "CICSTCPIPService",
+
```
AndrewTwydell, line 65: This is not accurate. The comment and constant name suggests definitions, but the actual resource name suggests the installed resource. Which do you need here?

packages/vsce/src/utils/resourceUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export async function runGetResource({ session, resourceName, regionName, cicsPlex, params }: {
+  session: Session,
+  resourceName: string,
+  regionName?: string,
+  cicsPlex?: string,
+  params?: { criteria?: string, parameter?: string; queryParams?: IResourceQueryParams; };
+}) {
+  let count = 0;
+  while (count <= 1) {
```
AndrewTwydell, line 31: Maybe changing `count` to be a more readable boolean could be nice, although it'll achieve the same result. 

I also wonder if there's a way we can achieve this without a while loop, perhaps moving the functionality inside the while to another method, and calling it from the try block AND the catch (if the 401 is received). Perhaps unecessary change at this point though - depends how strongly others feel about a while loop!

AndrewTwydell (changes requested):
Few changes, will also test functionality locally

packages/vsce/src/commands/disableCommands/disableTransactionCommand.ts
```diff
@@ -109,23 +110,19 @@ function disableTransaction(
   session: imperative.AbstractSession,
   parms: ICommandParams
 ): Promise<ICMCIApiResponse> {
-  const requestBody: any = {
+  return runPutResource({
+    session: session,
+    resourceName: CicsCmciConstants.CICS_LOCAL_TRANSACTION,
+    cicsPlex: parms.cicsPlex,
+    regionName: parms.regionName,
+    params: {"criteria": `PROGRAM='${parms.name}'`}
```
AndrewTwydell, line 118:
```suggestion
    params: {"criteria": `TRANID='${parms.name}'`}
```

packages/vsce/src/commands/openLocalFileCommand.ts
```diff
@@ -104,23 +105,19 @@ export function getOpenLocalFileCommand(tree: CICSTree, treeview: TreeView<any>)
 }
 
 function openLocalFile(session: imperative.AbstractSession, parms: ICommandParams): Promise<ICMCIApiResponse> {
-  const requestBody: any = {
+  return runPutResource({
+    session: session,
+    resourceName: CicsCmciConstants.CICS_PROGRAM_RESOURCE,
```
AndrewTwydell, line 110:
```suggestion
    resourceName: CicsCmciConstants.CICS_CMCI_LOCAL_FILE,
```

AndrewTwydell (changes requested):
Couple things to tweak - and I need a longer, more thorough, look at CICSTree. There's another `while` loop in there which I'm keen to avoid but not absolutely necessary. Will test locally and we can go over it tomorrow.

AndrewTwydell (approved):
Wahoo! LGTM - tested thoroughly locally and seems all good. Nice to be able to use tokens and not rely on basic auth for every request! 👍🏼

davenice (approved):
This looks good to me! I know you and Andrew have tested it pretty hard.

Let's pause and check whether @zFernand0 wants to give another review.

(BTW interesting point from @zFernand0 about the CLI in daemon mode. I've not tried that out. It's certainly a potential use case for making use of tokens in the future.)

packages/vsce/src/utils/resourceUtils.ts
zFernand0, line 1: I really like the way this was refactored. Very clever!  🧠

zFernand0 (approved):
LGTM! 😋 
---

This is really cool!  🎉 
One step closer to APIML integration 😋

Issue #227, opened 2025-02-06
Duplicate plexes exist when connecting to a multi-cmas system


AndrewTwydell:
**Describe the bug**

When connecting to a multi-cmas system, duplicate records are returned from CMCI for a single plex as multiple CMASs know about it. This results in the same plex appearing twice under the session.

zFernand0:
- resolved in #228

PR #228, opened 2025-02-07
Removed duplicate CICSplexes so they don't appear in the tree

enamkhan:
Filter duplicate cicsplexes so that these do not appear in the tree.

**How to Test**
Have added unit tests for the compare function,

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
This is for defect https://github.com/zowe/cics-for-zowe-client/issues/227.

### DIFF
--- a/packages/vsce/src/utils/plexUtils.ts
+++ b/packages/vsce/src/utils/plexUtils.ts
@@ -9,8 +9,6 @@
  *
  */
 
-import { string } from "yargs";
-
 export interface ICicsPlexInfo {
   _keydata: string;
   accesstype: string,
--- a/packages/vsce/src/utils/plexUtils.ts
+++ b/packages/vsce/src/utils/plexUtils.ts
@@ -20,7 +20,7 @@ export function filterCicsplexByConstraints(cicscicsplex: any[]) {
 
   for (const plex of cicscicsplex) {
     const plexname: string = plex["plexname"];
-    let cicsplex = allcicsplexes.get(plexname);
+    const cicsplex = allcicsplexes.get(plexname);
     if (!cicsplex) {
       allcicsplexes.set(plexname, plex);
     }
--- a/packages/vsce/src/utils/plexUtils.ts
+++ b/packages/vsce/src/utils/plexUtils.ts
@@ -26,20 +26,17 @@ export interface ICicsPlexInfo {
   updaters: string
 }
 
-export function evaluateCicsPlex(plex: ICicsPlexInfo): number {
+export function scoreCicsPlexByStatus(plex: ICicsPlexInfo): number {
   return  (plex.status === "ACTIVE" && 7) + (plex.accesstype === "LOCAL" && 5) + (plex.mpstatus === "YES" && 3);
 }
 
 // pick the highest scoring cicsplexes if there are duplicates
-export function filterCicsplexByConstraints(cicscicsplex: ICicsPlexInfo[]) {
+export function getBestCICSplexes(cicscicsplex: ICicsPlexInfo[]) {
   const allcicsplexes = new Map<string, ICicsPlexInfo>();
-  cicscicsplex.sort((a, b) => evaluateCicsPlex(b) - evaluateCicsPlex(a));
+  cicscicsplex.sort((a, b) => scoreCicsPlexByStatus(a) - scoreCicsPlexByStatus(b));
 
   for (const plex of cicscicsplex) {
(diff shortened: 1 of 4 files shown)


packages/vsce/src/utils/plexUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (old) {
+      if (compareCicsplexes(plex, old) > 0) {
+        allcicsplexes.set(plexname, plex);
+      }
+    } else {
+      allcicsplexes.set(plexname, plex);
+    }
+  }
+  return allcicsplexes;
+}
```
AndrewTwydell, line 46: A few things I'd like to discuss here.

Firstly, I think we could name things slightly clearer to make it more readable and understandable. (eg. `allcicsplexes` -> `uniqueCICSplexes`, or `compareCicsplexes` -> `isMoreRelevantPlex`)

Secondly, baring in mind we're essentially comparing variables to be either the same or not (which is a yes or no), is there a reason we're converting booleans to numbers and back again?? How about something like...

I also would like to try and avoid adding more `any` types - we have enough already! While we don't have a plex resource interface available in the SDK, we could define it locally or in a different file (these interfaces are in a WIP PR).

Any benefit in moving the comparison method out so we can unit test it separately??

```suggestion
type cicscicsplex = { accesstype: string, mpstatus: string, plexname: string, status: string; };

const boolsAreDifferent = function (val1: boolean, val2: boolean): boolean {
  return val1 !== val2;
};

export function isMoreRelevantPlex(plex1: cicscicsplex, plex2: cicscicsplex): boolean {
  if (boolsAreDifferent(plex1.status === "ACTIVE", plex2.status === "ACTIVE")) {
    return true;
  }
  if (boolsAreDifferent(plex1.mpstatus === "YES", plex2.mpstatus == "YES")) {
    return true;
  }
  return boolsAreDifferent(plex1.accesstype === "LOCAL", plex2.accesstype === "LOCAL");
}

export function filterCicsplexByConstraints(cicsplexRecords: cicscicsplex[]): Map<string, cicscicsplex> {

  const uniqueCICSplexes = new Map<string, cicscicsplex>();

  /**
   * Loop each plex record. 
   * If we're yet to see a plex with that name, or we found a more relevant plex, replace it.
   */
  for (const plexRecord of cicsplexRecords) {

    const plexName = plexRecord.plexname;
    const plexAlreadyFound = uniqueCICSplexes.has(plexName);

    if (
      !plexAlreadyFound || (plexAlreadyFound && isMoreRelevantPlex(plexRecord, uniqueCICSplexes.get(plexName)))
    ) {
      uniqueCICSplexes.set(plexName, plexRecord);
    }
  }
  return uniqueCICSplexes;
}
```

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -216,13 +218,15 @@ export class ProfileManagement {
...
+
+          cicscicsplexs.forEach((value: {}, key: string) => {
             infoLoaded.push({
-              plexname: plex.plexname,
+              plexname: key,
               regions: [],
               group: false,
             });
-          }
+          });
```
AndrewTwydell, line 229: To loop only the keys, you could do something like
```
        [...cicscicsplexs.keys()].forEach((plexName: string) => {
            infoLoaded.push({
              plexname: plexName,
              regions: [],
              group: false,
            });
          });
```

Achieves the same though, so not crucial.

AndrewTwydell (changes requested):
Much better!! Very much appreciated.

Couple comments, mainly around typing, then we're good to go!

packages/vsce/src/utils/plexUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+export interface ICicsPlexInfo {
```
AndrewTwydell, line 12: Just for future reference, this is temporary and will be replaced by an SDK interface.

AndrewTwydell (approved):
Great, thanks Enam.

This has been tested against real systems facing this issue to see it in action.

AndrewTwydell (approved):

davenice (approved):

Issue #56, opened 2022-03-10
Navigate from CICS region to the CEEMSG log


Joe-Winchester:
For developers who use the `DISPLAY` statement for debugging programs, this ends up being written to the `CEEMSG` JES file for the CICS job.

The request is to provide a point and shoot solution to go from a CICS region to open the CEEMSG file.

Issues to solve:
- How to know the right z/OSMF, FTP, RSE (i.e. base profile) is associated with the CICS profile.
- Existing analogous things are submit job and navigate to job, however this just goes from tree view to tree view.  For CICS it may be possible to remove the need for a user to even open the JES view and just go directly to open the CEEMSG file.

ruc.dk

Joe-Winchester:
Hi @t1m0thyj ,    I love the idea of `profileLinks` and would like to find out more.   Do you have a design doc link, or maybe a scrum recording I could look at,  or perhaps find some time to just chat 1:1 and I can show you the sort of scenario we were thinking to navigate between different VS Code plugins with related data (that would also cover some of the Db2 plugins) and it'd be great to see if `profileLinks` are going to be able to help ?

t1m0thyj:
@Joe-Winchester Here is a design doc link: https://ent.box.com/s/thqbdjbz9rd2heohxb2ds27kv4sdlgv3

Feel free to reach out on Slack, would be happy to take some time to chat 1:1 to better understand your use case 🙂 

I'm curious to know your opinion about which direction links should be defined. For example if you want to associate a z/OSMF profile with a CICS profile, which of the following `profileLinks` definitions seems more intuitive:
```json
  "profileLinks": {
    "lpar1_zosmf": ["lpar1_cics"]
  }
```
or
```json
  "profileLinks": {
    "lpar1_cics": ["lpar1_zosmf"]
  }
```

SwatiMohapatra:
Hey @t1m0thyj , the design box link which you sent seems to be unavailable for me. Can you resend it please?

PR #237, opened 2025-02-24
show region logs

davenice:
**What It Does**
Addresses https://github.com/zowe/cics-for-zowe-client/issues/56

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog
- [x] updated the readme

**Additional Comments**


packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  // filter out profiles that are not in the same base as the cics profile
+  const matchingProfiles: IProfileLoaded[] = [];
+  if (baseForCicsProfile) {
+    for (const profile of zosProfiles) {
+      if (baseForCicsProfile && baseForCicsProfile?.name === (await ProfileManagement.getProfilesCache().fetchBaseProfile(profile.name))?.name) {
+        matchingProfiles.push(profile);
+      }
+    }
+  }
```
davenice, line 35: @JillieBeanSim / @zFernand0 - this is the part of the PR I'm particularly interested in your thoughts on.

I'm taking a CICS profile and a list of proposed zosProfiles.

I want to identify a zosProfile that matches the CICS profile. I'm currently doing this by requesting the base profile for each proposed zosProfile to see if it matches the base profile of the cicsProfile.

Is this the best way, currently?

packages/vsce/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Add context menu on a region to show the region's logs [#56](https://github.com/zowe/cics-for-zowe-client/issues/56)
```
```suggestion
- Enhancement: Add context menu on a region to reveal the region's logs in the jobs tree. [#56](https://github.com/zowe/cics-for-zowe-client/issues/56)
```

packages/vsce/package.json
```diff
@@ -334,6 +334,11 @@
         "title": "Show Attributes",
         "category": "IBM CICS for Zowe Explorer"
       },
+      {
+        "command": "cics-extension-for-zowe.showRegionLogs",
+        "title": "Show Logs",
```
AndrewTwydell, line 339:
```suggestion
        "title": "Show region's logs",
```

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+      // unlikely to get here but logging this would be useful in future
+    }
+  }
+  return jobid;
+}
+
+export function getShowRegionLogs(treeview: TreeView<any>) {
+  return commands.registerCommand("cics-extension-for-zowe.showRegionLogs", async (node) => {
+    const allSelectedRegions = findSelectedNodes(treeview, CICSRegionTree, node);
+    if (!allSelectedRegions || !(allSelectedRegions.length == 1)) {
```
AndrewTwydell, line 91: You can keep it this way, and change the `==` to `===`, or you can do something like this to check it's NOT 1
```suggestion
    if (!allSelectedRegions || allSelectedRegions.length !== 1) {
```

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+      await window.showErrorMessage("One CICS region must be selected");
+      return;
+    }
+    const selectedRegion: CICSRegionTree = allSelectedRegions[0];
+    const zosProfiles = (await ProfileManagement.getProfilesCache().fetchAllProfiles()).filter(
+      (profile) => profile.type === "zosmf" || profile.type === "rse",
```
AndrewTwydell, line 97: By no means mandatory, but I like doing something like this when checking if a value is 1 of multiple options.

```suggestion
      (profile) => ["zosmf", "rse"].includes(profile.type),
```

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // find profiles that match by base profile or hostname
+    const matchingZosProfiles = await findRelatedZosProfiles(selectedRegion.parentSession.profile, zosProfiles);
+    if (matchingZosProfiles) {
+      chosenProfileName = matchingZosProfiles.name;
+    } else {
+      // we couldn't find a matching profile - prompt the user with all zos profiles
+      chosenProfileName = await promptUserForProfile(zosProfiles);
+      if (chosenProfileName === null) {
+        window.showErrorMessage("Could not find any z/OSMF or RSE profiles.");
```
AndrewTwydell, line 110: It's possible this code can be reached if the user is prompted to choose a profile and presses `Esc` so the error message isn't too relevant (I think). We could see what behaviour we do in other places where quickpicks can be cancelled.

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+export async function getJobIdForRegion(selectedRegion: CICSRegionTree): Promise<string> {
+  // when we have a CICSRGN table we have jobid, but not when we have a
+  // MAS table. we get either CICSRGN or MAS depending on where we are in the
+  // tree. request CICSRGN if jobid isn't available.
+  let jobid: string = selectedRegion.region.jobid;
+  if (!jobid) {
+    let response: ICMCIApiResponse;
+    try {
+      response = await getResource(selectedRegion.parentSession.getSession(), {
+        name: "CICSRegion",
```
AndrewTwydell, line 75: We have a constant for this - and if we don't, we should!

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+        cicsPlex: selectedRegion.parentPlex.plexName,
+        regionName: selectedRegion.region.cicsname,
+      });
+      jobid = response.response.records.cicsregion.jobid;
```
AndrewTwydell, line 79: Something to be cautious of (although you should be fine here as the hope is you'd only ever get one record back)... getResource's `response.records.cicsregion` can be either an array of objects (each object being a region record) or an object itself. This is silly, but that's why we have `toArray(...)` dotted around. Something to bare in mind but should be fine here.

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+  // filter down to just profiles that have the same hostname as our cics connection
+  const cicsHostProfiles = zosProfiles.filter((profile) => cicsProfile.profile.host === profile.profile.host);
+
+  return cicsHostProfiles[0];
+}
+
+function promptUserForProfile(zosProfiles: IProfileLoaded[]): Thenable<string> {
+  const profileNames = zosProfiles.map((value) => {
+    return value.name;
+  });
```
AndrewTwydell, line 52:
```suggestion
  const profileNames = zosProfiles.map((value) => value.name);
```

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  // sort profiles with zosmf ones first to make zosmf the default
+  // also filter so we only automatically pick z/os connections that have credentials associated
+  zosProfiles = zosProfiles.sort((prof) => (prof.profile.type === "zosmf" ? -1 : 1)).filter((prof) => prof.profile.user);
+
+  // filter out profiles that are not in the same base as the cics profile
+  const matchingProfiles: IProfileLoaded[] = [];
+  if (baseForCicsProfile) {
+    for (const profile of zosProfiles) {
+      if (baseForCicsProfile && baseForCicsProfile?.name === (await ProfileManagement.getProfilesCache().fetchBaseProfile(profile.name))?.name) {
```
AndrewTwydell, line 32: Asserted it exists earlier in the line

```suggestion
      if (baseForCicsProfile && baseForCicsProfile.name === (await ProfileManagement.getProfilesCache().fetchBaseProfile(profile.name))?.name) {
```

AndrewTwydell (commented):
Looking good! Only some suggestions because you asked so nicely for some 😄

packages/vsce/src/commands/showLogsCommand.ts
davenice: good shout - will be in the next push.

davenice:
Here is a quick vid showing what this does and why it's not exactly what @Joe-Winchester suggested in the original issue :-)

Happy, @Joe-Winchester ?

https://github.com/user-attachments/assets/bb84aaf5-ca2c-4261-98b9-e6ef4aa51c2b

packages/vsce/src/commands/showLogsCommand.ts
davenice: I tested this - thanks for the thought. If you hit escape, the answer is undefined, which isn't null.

So the error message doesn't show, which is good. The behaviour I think is bang on what we'd want. However the undefined vs null seems a bit "luck rather than judgement" ... I'm thinking perhaps I add a comment to the method explaining the possible returns 🤔

packages/vsce/package.json
davenice: I think I could go to "Show Region Logs" but everything's capitalised in those context menus and I'm not keen on the apostrophe (although with no good reason!)

davenice: I've updated to `Show Region Logs`

packages/vsce/src/commands/showLogsCommand.ts
JillieBeanSim: For Profiles that match a particular CICS profile you may want to use [ProfilesCache.fetchAllProfiles()](https://github.com/zowe/zowe-explorer-vscode/blob/a05967b41a770e0ec2542a7b686c3a1c0e506b03/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts#L278) then filter down to the profiles with matching host.  Also, depending on the scenario you may want to filter even further on profile types you may want to support removing types like TSO, SSH or what doesn't fit.  All these profiles will already have any merging done if needed, relying on base profile can prove to be troublesome since users configure their Zowe Config in different ways and sometimes without base profiles.

JillieBeanSim: if you know the profile type there is also [ProfilesCache.fetchAllProfilesByType(type)](https://github.com/zowe/zowe-explorer-vscode/blob/a05967b41a770e0ec2542a7b686c3a1c0e506b03/packages/zowe-explorer-api/src/profiles/ProfilesCache.ts#L259) returning complete/merged profiles

davenice:
> Q: For the 3rd case where you prompt the user, is that in way stored/remembered anywhere or it's just a one-off ephemeral nativation ?

It's not stored anywhere. I did wonder about going deeper like if a connection didn't work correctly offering the prompt, and remembering what the user wanted. However I think by picking ones that have a user associated we will be fairly sure they work. And I think in most situations the user should sort the nesting or hostnames so things work! Could be an improvement for the future though :-)

> Q2: You're choosing `zOSMF` and `rse` as these are the only two connection types that support showing spool for jobs with `ACTIVE` status (which is why you are filtering away `ftp`). What if a new type comes along that can support showing spool for `ACTIVE` jobs ? I know the https://github.com/zowe/zowe-native-proto project is creating an SSH back end to support Zowe Explorer front end for MVS/USS/JES - not sure where this is with its support for `ACTIVE` spool but it could get there sometime soon. Is there a way to ping interface types to see if the profile type in a way that new ones can be introduced, or maybe I've over thinking and these events are so rare that if a user asked you'd just update the array to include it ?

Yes that's right - I didn't find any capability-based filtering that would have allowed that. The FTP connection, for instance, just returns nothing if asked for a running job's spool files so it's difficult to tell what each connection supports. I'm not even sure that I could find a 'category' that I could filter on to get everything that says it might fulfil the Zowe Explorer jobs view requirements.

I can't see a good option currently. 🤔

davenice:
The simplest (and not brilliant) option is that we could make the list a user preference if required in the future. I'm not all that inclined right now though - it's so easy to fix if we need to in the future.

AndrewTwydell:
The only fully seamless approach I can think of would to have a way within the profiles ecosystem to query registered profile types by action - something like "give me all the profiles (or profile types) that can list jobs". That would require some new profile metadata on ZE's end to know what each registered profile type can do, including extenders like FTP, SSH etc. 

An interesting point though. What would the outcome be if we listed ALL profiles (potentially filtering out CICS)? On the one hand, we would show all options, particularly if a user is using a custom registered profile type or a type we've not yet implemented. On the other hand, we'd have to error-handle the user picking a profile type NOT supported which will be increasingly more common and annoying 🤔 

I definitely think zOSMF and RSE will very much do to get this in... Just brain dumping 😄

davenice:
> What would the outcome be if we listed ALL profiles (potentially filtering out CICS)? On the one hand, we would show all options, particularly if a user is using a custom registered profile type or a type we've not yet implemented. On the other hand, we'd have to error-handle the user picking a profile type NOT supported which will be increasingly more common and annoying 🤔

I think we'd pick up the IA connection and any other Zowe extenders! 👎

Joe-Winchester:
> > What would the outcome be if we listed ALL profiles (potentially filtering out CICS)? On the one hand, we would show all options, particularly if a user is using a custom registered profile type or a type we've not yet implemented. On the other hand, we'd have to error-handle the user picking a profile type NOT supported which will be increasingly more common and annoying 🤔
> 
> I think we'd pick up the IA connection and any other Zowe extenders! 👎

Right, filtering all profiles without CICS runs the risk of picking up others that have nothing to do with base function.
The Zowe Explorer UI for data sets/USS/JES does have something along the right lines though, as when you select "Issue TSO Command" or "Issue MVS Command" it only shows you the profiles for z/OSMF and RSE (which are the only ones that can issue TSO/MVS commands through).  It doesn't include FTP in the list (as far as I can tell doing a quick smoke test)

image

Also, in a similar vein, when you issue "Issue Unix Command" against a z/OSMF profile it prompts for the ssh type profile (if you're against a z/OSMF connection), but if you're against a type of rse (that has a REST API to run USS commands) then there is no prompt that appears to ask the user to select ssh, which feels to me as though the `Issue UNIX Command` is able to do some profile capability querying.  

image

davenice:
Taking Joe's comments on board I've updated the implementation to check for connections that support JES function, filtering out FTP.

I think this is a reasonable balance of effort - it seems unlikely that we'll get another connection that supports JES but doesn't happen to support active jobs. Adding in a really fine-grained capability support doesn't seem fair for this PR.

davenice:
@AndrewTwydell - you had some concerns about whether there were a lot of scenarios where credentials wouldn't be present or the profile wouldn't work out of the box, based on your testing. Please do give it a bash!

AndrewTwydell:
When credentials are provided for the ZOSMF profile, things work great. It can decipher which profile it should use, gets credentials either from the keychain or plaintext in the config file and reveals the job in the tree - nice! 

When no credentials are provided, we have a couple of issues.
1. When the CICS and ZOSMF profiles share a parent, it offers a selection of ZOSMF profiles to choose from in a quickpick. Once chosen, you get a `Required session must be defined` error.
2. When they do NOT share a parent (I've got the profiles at a top level, this may be different if they are nested but not sharing a parent), you get a `Error running command cics-extension-for-zowe.showRegionLogs: Failed to load schema for profile type base.` error - my guess is you're trying to get the base profile for the given profile, which doesn't exist, and not validating that before trying to act on it.

davenice:
Thanks @AndrewTwydell.

(2) is dealt with - an error comes back from zowe explorer if you ask for the base profile of something without one. i've wrapped their method in one that will cope with an error.

(1) is trickier. I need to try and get a decent recreate and consult Billie.

davenice:
I think the code I've got here addresses your two issues, @AndrewTwydell.

However if we end up having to prompt the user for credentials, we get an error that the session isn't defined (I think the Zowe Explorer code is assuming we're prompting for a session whereas we don't have a session yet, we're looking at the profile).

This is still an improvement and arguably could be shippable, however I plan to carry on following up with the Zowe Explorer team whether we can achieve the end result without this error.

packages/vsce/__tests__/__utils__/globalMocks.ts
```diff
@@ -29,14 +29,14 @@ export const CICSProfileMock = {
...
-export const imperativeSession = new imperative.Session({
+export const imperativeSession = {
   user: "user",
   password: "pwd",
   hostname: "hostname",
   protocol: "https",
   type: "basic",
   rejectUnauthorized: false,
-});
+} as unknown as imperative.Session;
```
AndrewTwydell, line 39: I'm sure there was a reason for this, but it seems odd! Why can't we have the Session type annotation?

davenice: I cannot fathom why, but I get an error in the tests if I revert this change:
```
** FAIL   Zowe CICS VSCE Unit Tests  __tests__/__unit__/commands/showLogsCommand.unit.test.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'Session')

      30 | };
      31 |
    > 32 | export const imperativeSession = new imperative.Session({
         |                                                 ^
      33 |   user: "user",
      34 |   password: "pwd",
      35 |   hostname: "hostname",

      at Object.<anonymous> (__tests__/__utils__/globalMocks.ts:32:49)
      at Object.<anonymous> (__tests__/__unit__/commands/showLogsCommand.unit.test.ts:20:1)
**
```

Whereas with my change:
```
 PASS   Zowe CICS VSCE Unit Tests  __tests__/__unit__/commands/showLogsCommand.unit.test.ts
```

Any ideas?

packages/vsce/package.json
```diff
@@ -453,6 +458,11 @@
           "command": "cics-extension-for-zowe.showRegionAttributes",
           "group": ""
         },
+        {
+          "when": "view == cics-view && viewItem =~ /cicsregion.*\\.active/",
```
AndrewTwydell, line 462: Could do this if you don't want to be able to select multiple? It's what we do with manage profile.

```suggestion
          "when": "view == cics-view && viewItem =~ /cicsregion.*\\.active/ && !listMultiSelection",
```

AndrewTwydell (approved):
Few comments but looks good. Tested locally and fixed those issues I found before 👍🏼

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+export async function getJobIdForRegion(selectedRegion: CICSRegionTree): Promise<string> {
+  // when we have a CICSRGN table we have jobid, but not when we have a
+  // MAS table. we get either CICSRGN or MAS depending on where we are in the
+  // tree. request CICSRGN if jobid isn't available.
+  let jobid: string = selectedRegion.region.jobid;
+  if (!jobid) {
+    try {
+      const { response } = await getResource(selectedRegion.parentSession.getSession(), {
```
enamkhan, line 100: Should be using runGetResource as that will handle ltpa token expiry.

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+    const allSelectedRegions = findSelectedNodes(treeview, CICSRegionTree, node);
+    if (!allSelectedRegions || allSelectedRegions.length !== 1) {
+      window.showErrorMessage("One CICS region must be selected");
+      return;
+    }
+
+    const selectedRegion: CICSRegionTree = allSelectedRegions[0];
+    const jobid = await getJobIdForRegion(selectedRegion);
+    if (!jobid) {
+      window.showErrorMessage(`Could not find Job ID for region ${selectedRegion.region.cicsname}.`);
```
enamkhan, line 136: We could add some logging as well now.

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+  } catch (ex) {
+    // profile does not support JES API
+  }
+  return false;
+}
+
+export function getShowRegionLogs(treeview: TreeView<any>) {
+  return commands.registerCommand("cics-extension-for-zowe.showRegionLogs", async (node) => {
+    const allSelectedRegions = findSelectedNodes(treeview, CICSRegionTree, node);
+    if (!allSelectedRegions || allSelectedRegions.length !== 1) {
```
enamkhan, line 128: Changing the when clause would make this redundant.

enamkhan (changes requested):
We should change the getResource call to make it consistent.

packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  if (matchingProfiles.length > 0) {
+    return matchingProfiles[0];
+  }
+
+  // we couldn't find anything within a profile group
+  // filter down to just profiles that have the same hostname as our cics connection
+  const cicsHostProfiles = zosProfiles.filter((profile) => cicsProfile.profile.host === profile.profile.host);
+
+  return cicsHostProfiles[0];
```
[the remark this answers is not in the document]
davenice: yes - if there's nothing in the array nothing bad happens so I decided not to add a specific check (all those checks out me as a Java developer, right? 😂)

davenice:
Here's an example of the logs. I've annoted with ***s as the start of lines
```
2025-03-28 23:37:21.665 [info] Initialized logger for IBM CICS for Zowe Explorer
2025-03-28 23:37:21.665 [info] IBM CICS for Zowe Explorer 3.5.0
2025-03-28 23:37:21.665 [info] IBM CICS for Zowe Explorer log level: Info

*** a normal request with the default info level
2025-03-28 23:37:47.889 [info] Calling zowe.jobs.setJobSpool for region IYCWEPL1: winmvs28.zosmf / STC03435

*** a request with debug level set where we have nested profiles
2025-03-28 23:37:53.421 [debug] Showing region logs for region IYCWEPM1
2025-03-28 23:37:53.422 [debug] GET request - Resource [CICSRegion], CICSplex [CICSEX62], Region [IYCWEPM1]
2025-03-28 23:37:53.517 [debug] Job ID for region: STC03434
2025-03-28 23:37:53.532 [debug] Located matching z/OS profile by base profile: winmvs28.zosmf
2025-03-28 23:37:53.535 [info] Calling zowe.jobs.setJobSpool for region IYCWEPM1: winmvs28.zosmf / STC03434

*** a request with debug level set where we can't identify a z/OS Profile and need to prompt the user
2025-03-28 23:37:56.470 [debug] Showing region logs for region IYK2ZOD1
2025-03-28 23:37:56.472 [debug] Job ID for region: JOB53533
2025-03-28 23:37:57.872 [debug] User picked z/OS profile: winmvs28.zosmf
2025-03-28 23:37:57.872 [info] Calling zowe.jobs.setJobSpool for region IYK2ZOD1: winmvs28.zosmf / JOB53533

*** a request with debug level set where we don't have credentials
2025-03-28 23:40:00.485 [debug] Showing region logs for region IYK2ZOD1
2025-03-28 23:40:00.489 [debug] Job ID for region: JOB53533
2025-03-28 23:40:12.947 [debug] Prompting for credentials for winmvs28.zosmf
2025-03-28 23:40:17.146 [debug] User picked z/OS profile: winmvs28.zosmf
2025-03-28 23:40:17.146 [info] Calling zowe.jobs.setJobSpool for region IYK2ZOD1: winmvs28.zosmf / JOB53533
```

davenice:
Here's an example of the logs. I've annoted with ***s as the start of lines
```
2025-03-28 23:37:21.665 [info] Initialized logger for IBM CICS for Zowe Explorer
2025-03-28 23:37:21.665 [info] IBM CICS for Zowe Explorer 3.5.0
2025-03-28 23:37:21.665 [info] IBM CICS for Zowe Explorer log level: Info

*** a normal request with the default info level
2025-03-28 23:37:47.889 [info] Calling zowe.jobs.setJobSpool for region IYCWEPL1: host.zosmf / STC03435

*** a request with debug level set where we have nested profiles
2025-03-28 23:37:53.421 [debug] Showing region logs for region IYCWEPM1
2025-03-28 23:37:53.422 [debug] GET request - Resource [CICSRegion], CICSplex [CICSEX62], Region [IYCWEPM1]
2025-03-28 23:37:53.517 [debug] Job ID for region: STC03434
2025-03-28 23:37:53.532 [debug] Located matching z/OS profile by base profile: host.zosmf
2025-03-28 23:37:53.535 [info] Calling zowe.jobs.setJobSpool for region IYCWEPM1: host.zosmf / STC03434

*** a request with debug level set where we can't identify a z/OS Profile and need to prompt the user
2025-03-28 23:37:56.470 [debug] Showing region logs for region IYK2ZOD1
2025-03-28 23:37:56.472 [debug] Job ID for region: JOB53533
2025-03-28 23:37:57.872 [debug] User picked z/OS profile: host.zosmf
2025-03-28 23:37:57.872 [info] Calling zowe.jobs.setJobSpool for region IYK2ZOD1: host.zosmf / JOB53533

*** a request with debug level set where we don't have credentials
2025-03-28 23:40:00.485 [debug] Showing region logs for region IYK2ZOD1
2025-03-28 23:40:00.489 [debug] Job ID for region: JOB53533
2025-03-28 23:40:12.947 [debug] Prompting for credentials for host.zosmf
2025-03-28 23:40:17.146 [debug] User picked z/OS profile: host.zosmf
2025-03-28 23:40:17.146 [info] Calling zowe.jobs.setJobSpool for region IYK2ZOD1: host.zosmf / JOB53533
```

davenice:
Right. Let's go again!

enamkhan (approved):
LGTM!

PR #247, opened 2025-03-13
adding wiremock configuration and adding the wiremock server to cics

Snehabayya:
…section

**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/__tests__/__e2e__/01_EditProjectTeamConfiguration.test.ts
+++ b/packages/vsce/__tests__/__e2e__/01_EditProjectTeamConfiguration.test.ts
@@ -11,7 +11,7 @@
 
 import { assert, expect } from "chai";
 import * as path from "path";
-import { ActivityBar, DefaultTreeSection, EditorView, InputBox, VSBrowser } from "vscode-extension-tester";
+import { ActivityBar, DefaultTreeSection, EditorView, InputBox, TextEditor, VSBrowser } from "vscode-extension-tester";
 
 describe("Create Project Level Team Configuration File Scenario", () => {
   let cicsTree: DefaultTreeSection;
@@ -86,6 +86,19 @@ describe("Create Project Level Team Configuration File Scenario", () => {
     // Check zowe.config.json was opened - could check content here
     expect(titles.some((title) => title.startsWith("zowe.config.json"))).is.true;
 
+    // Open the editor for zowe.config.json
+    const editor = await editorView.openEditor("zowe.config.json") as TextEditor;
+   
+    const isWmAvailable = await editor.getText()
+    if(!isWmAvailable.includes("wiremock_server")){
+      const wiremockConfig = ",\n\"wiremock_server\": {\n   \"type\": \"cics\",\n\"properties\": {\n     \"host\": \"localhost\",\n\"port\": 8080,\n\"rejectUnauthorized\": false,\n\"protocol\": \"http\"\n}\n}"
+    
+      // add the wiremock configuration at the given coordinates
+      await editor.typeTextAt(47,10,wiremockConfig);
+      await editor.save();
+      const editorText = await editor.getText();
+      expect(editorText).to.contain("wiremock_server");
+    }
     cicsTree.takeScreenshot();
   });
 });
--- a/packages/vsce/__tests__/__e2e__/02_AddWiremockConfigurationProjectLevel.test.ts
+++ b/packages/vsce/__tests__/__e2e__/02_AddWiremockConfigurationProjectLevel.test.ts
@@ -1,0 +1,62 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { assert, expect } from "chai";
+import { ActivityBar, DefaultTreeSection, EditorView, InputBox, TreeItem, VSBrowser } from "vscode-extension-tester";
+
+describe("Create Project Level WireMock File Scenario", () => {
+  let cicsTree: DefaultTreeSection;
+  let quickPick: InputBox;
+  let editorView: EditorView;
+
+  before(async function () {
+    // switch to the cics extension view
+    const zoweExplorer = await new ActivityBar().getViewControl("Zowe Explorer");
+    assert(zoweExplorer !== undefined);
+    const view = await zoweExplorer.openView();
+    cicsTree = await view.getContent().getSection("cics");
+
+    // click the plus icon in cics
+    await cicsTree.click();
+    await cicsTree.expand();
+    const plusIcon = await cicsTree.getAction(`Create a CICS Profile`);
+    assert(typeof plusIcon !== "undefined");
+    await plusIcon.click();
+  });
+
+  it("Quick Pick Option test", async () => {
+    // Find quickpick
+    quickPick = await InputBox.create();
+    const qpItems = await quickPick.getQuickPicks();
+    const qpLen = qpItems.length;
+
+    // selecting wiremock_server from the quickpick
+    for(let i=0; i<qpLen; i++){
+        const wiremocklabel = await qpItems[i].getLabel();
+        if(wiremocklabel.includes("wiremock_server")){
+            expect(wiremocklabel).contains("wiremock_server")
+            await quickPick.selectQuickPick(i);
+            break;
+        }
+    }
+    cicsTree.takeScreenshot();
+  });
+
+  it("WireMock Server available test", async ()=>{
+    // checking if wiremock_server is available in the cics section
+    const potentialWireMockServer = await cicsTree.findItem("wiremock_server");
+    assert(typeof potentialWireMockServer !== "undefined");
+    const wireMockServer = potentialWireMockServer;
+    expect(wireMockServer).exist;
+    wireMockServer.click();
+    cicsTree.takeScreenshot();
+  });
+});
--- a/packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
+++ b/packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
@@ -44,7 +44,16 @@
                 "host": "localhost",
                 "rejectUnauthorized": true
             }
-        }
+        },
+        "wiremock_server": {
+           "type": "cics",
+           "properties": {
+                "host": "localhost",
+                "port": 8080,
+                "rejectUnauthorized": false,
+                "protocol": "http"
+                }
+                }
   },
     "defaults": {
         "zosmf": "zosmf",


packages/vsce/__tests__/__e2e__/01_EditProjectTeamConfiguration.test.ts
```diff
@@ -86,6 +86,19 @@ describe("Create Project Level Team Configuration File Scenario", () => {
...
+    const isWmAvailable = await editor.getText()
+    if(!isWmAvailable.includes("wiremock_server")){
+      const wiremockConfig = ",\n\"wiremock_server\": {\n   \"type\": \"cics\",\n\"properties\": {\n     \"host\": \"localhost\",\n\"port\": 8080,\n\"rejectUnauthorized\": false,\n\"protocol\": \"http\"\n}\n}"
+    
+      // add the wiremock configuration at the given coordinates
+      await editor.typeTextAt(47,10,wiremockConfig);
+      await editor.save();
+      const editorText = await editor.getText();
+      expect(editorText).to.contain("wiremock_server");
+    }
```
AndrewTwydell, line 101: I really like how you've looked into how to do this, and this will definitely come in useful!

However, because you've added the `wiremock_server` profile to the zowe.config.json we are using in this test, this will never executed right? I also believe it's good practice to know the environment the tests are running in (knowing what profiles are present so we can accurately test them) so I'm not sure we need this uncertainty. 

Perhaps an alternative is tweaking this to open the editor and reading the content still, but changing it from adding the profile if it's not there, to expecting it to be there. Something like...

```suggestion
    expect(isWmAvailable.includes("wiremock_server")).to.be.true;
```

packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
```diff
@@ -44,7 +44,16 @@
                 "host": "localhost",
                 "rejectUnauthorized": true
             }
-        }
```
almightyrush, line 47: The config file should be reseted  to the original after the test is completed. This will make sure that `if(!isWmAvailable.includes("wiremock_server"))` will always return true and adding the profile can only be done from the test.

hint: use `after()`

packages/vsce/__tests__/__e2e__/02_AddWiremockConfigurationProjectLevel.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { assert, expect } from "chai";
+import { ActivityBar, DefaultTreeSection, EditorView, InputBox, TreeItem, VSBrowser } from "vscode-extension-tester";
+
+describe("Create Project Level WireMock File Scenario", () => {
+  let cicsTree: DefaultTreeSection;
+  let quickPick: InputBox;
+  let editorView: EditorView;
+
+  before(async function () {
```
almightyrush, line 20: Use shorthand version 
`before(async () => {})`

packages/vsce/__tests__/__e2e__/01_EditProjectTeamConfiguration.test.ts
davenice: i was wondering about this too - i think the intention is to add the wiremock_server info into the the file at runtime ... but specifying the exact point in the file is brittle - likely to break if we update the format of that file at all.

i'm not sure exactly what the solution is...

packages/vsce/__tests__/__e2e__/02_AddWiremockConfigurationProjectLevel.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const qpLen = qpItems.length;
+
+    // selecting wiremock_server from the quickpick
+    for(let i=0; i<qpLen; i++){
+        const wiremocklabel = await qpItems[i].getLabel();
+        if(wiremocklabel.includes("wiremock_server")){
+            expect(wiremocklabel).contains("wiremock_server")
+            await quickPick.selectQuickPick(i);
+            break;
+        }
```
davenice, line 48: a tiny tweak that would help us diagnose things in the future - i think before we finish this for loop we should fail if we haven't found the wiremock_server entry?

that way rather than waiting to fail in the next section when we haven't got it in the list of profiles, we would find out that the problem was it wasn't in the picker of possible profiles.

does that make sense?

packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
davenice: perhaps we call this zowe.config.json.original and then copy it to zowe.config.json at the start of each test?

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        
+            // Selecting wiremock_server from the quickpick
+            let i, wiremockLabel;
+            for(i=0; i<qpLen; i++){
+                wiremockLabel = await qpItems[i].getLabel();
+                if(wiremockLabel.includes(text)){
+                    break;
+                }
+            }
+            expect(wiremockLabel).contains(text);
```
davenice, line 77: i like that we're checking we have the right thing here (and not just the last one!) 👍 

if this fails because we don't manage to write the correct config file correctly and therefore the profile isn't available, will we get an error message that helps us works that out? it's really good when you can read through the log of failed unit tests and just _know_ roughly what went wrong without having to dig into the code!

Perhaps the it("...") text could be something like "Add the wiremock CICS profile to the tree using the create profile toolbar option".

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+        it("Wiremock Profile Available Under CICS Section Test", async ()=>{
+            // Checking if wiremock_server profile is available under the cics section
+            const potentialWireMockServer: TreeItem | undefined  = await cicsTree.findItem(text);
+            wiremockServer = potentialWireMockServer;
+            expect(potentialWireMockServer).exist;
+            cicsTree.takeScreenshot();
+          });
+
+        it("Should List The Regions Under Wiremock Profile", async () => {
```
davenice, line 92: i think these are well named 👍 thanks!

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,116 @@
...
+            const wmLen = wmItems?.length;
+            expect(wmLen).equals(2);
+
+            const plex1 = await wmItems?.at(0)?.getLabel();
+            expect(plex1).contains("CICSEX61");
+
+            const plex2 = await wmItems?.at(1)?.getLabel();
+            expect(plex2).contains("DUMMY907");
+            cicsTree.takeScreenshot();
+          }).timeout(5000);
```
davenice, line 114: were we going to go for setting a default timeout so we didn't need to set a timeout on each test? or does this test need a slightly longer one?

packages/vsce/__tests__/__e2e__/09_RemoveWiremockProfile.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { expect } from "chai";
+import { EditorView, TextEditor } from "vscode-extension-tester";
+import { restoreOriginalProfile} from "./e2e_globalMocks";
+
+describe("Remove The Wiremock Profile From The Configuration File", async() => {
+    let editorView: EditorView;
+    const text = "wiremock_server";
+
+    it("Remove Wiremock Profile", async () => {
```
davenice, line 20: I'm wondering if we should have a test next that checks that it's disappeared from the tree and isn't available in the quick pick, if that's fairly quick to do.

That could be in the next PR though, doesn't have to be in this one!

packages/vsce/__tests__/__e2e__/e2e_globalMocks.ts
```diff
@@ -0,0 +1,51 @@
...
+    jsonOriginalProfile = {...jsonFile.profiles};
+    const newProfile = {
+        ...jsonFile.profiles,
+        ...wiremock_profile
+    };
+    jsonFile.profiles = newProfile;
+    fs.writeFileSync(jsonFilePath, JSON.stringify(jsonFile, null, 2), 'utf8');
+};
+
+export function restoreOriginalProfile(): void {
```
davenice, line 41: similarly i think restoreOriginalConfigFile() might be better - we're putting the config file back to how it was.

davenice (changes requested):
Hey @Snehabayya - this looks good a couple of minor wording suggestions

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
Snehabayya: We don't get any error message apart from the assertion error which says that the test has failed because of the mismatch of the names. I thought maybe this would be enough but if needed I can print the error message to the screen.

Yup got it, I will change the test name

packages/vsce/__tests__/__e2e__/09_RemoveWiremockProfile.test.ts
Snehabayya: Yeah sure I will add that

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
davenice: no i don't think you need to print the error message to the screen - improving the test name will be enough!

if i see that "Quick Pick Option test" failed, i don't really know what that means, whereas if i see "Add the wiremock CICS profile to the tree using the create profile toolbar option" i do.

I think that will be sufficient :-D

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
Snehabayya: This test needs a longer timeout because at line 96 we are using sleep method. So we are manually overwriting the default timeout which is 2 sec. 

We are using sleep method because the test is happening so fast and we are unable to get the children of the treeitem in the default timeout

davenice: Great, that sounds reasonable. Thanks!

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,116 @@
...
+            wiremockServer = potentialWireMockServer;
+            expect(potentialWireMockServer).exist;
+            cicsTree.takeScreenshot();
+          });
+
+        it("Should List The Regions Under Wiremock Profile", async () => {
+            // Expand the wiremock_server profile and list the regions under it
+            await wiremockServer?.click();
+            await wiremockServer?.collapse();
+            await sleep(3000);
```
davenice, line 96: I was thinking about this over the weekend (sad, I know!)

It's fine to finish and merge this PR as it is, but I think on reflection we need to avoid sleeps. That's because these tests will run in line with the build - we'll only need to add 10x 3 second sleeps across our tests and the whole build will take 30 seconds longer than it potentially needs to.

So I think that in a follow-up PR it would be good to work out a way of instead of saying "wait for 3 seconds", we say "wait until x has happened", probably providing an anonymous function that checks that the wiremockServer node doesn't have any children or similar.

Do you think this is feasible - @almightyrush / @AndrewTwydell ? If we work this out nicely now we can reuse the same technique for future tests!

davenice: Sounds good - a 100ms sleep is better for now - but as a strategy, avoiding sleeps will be great 👌

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { assert, expect } from "chai";
+import { ActivityBar, DefaultTreeSection, EditorView, InputBox, TreeItem, TextEditor, ViewPanelAction } from "vscode-extension-tester";
+import { addWiremockProfileToConfigFile, sleep } from "./e2e_globalMocks";
+
+describe("Test Suite For Adding Wiremock Profile And Listing The Regions In CICS", () => {
```
davenice, line 16:
```suggestion
describe("Test Suite For Adding Wiremock Profile And Listing The CICSplexes", () => {
```

Last thing - I realised that so far we're only listing the plexes, not the regions - so updating this description!

packages/vsce/__tests__/__e2e__/02_AddWiremockProfileAndListRegionsInCics.test.ts
```diff
@@ -0,0 +1,115 @@
...
+          });
+
+        it("Should Display Wiremock Profile Under CICS Section", async ()=>{
+            // Checking if wiremock_server profile is available under the cics section
+            wiremockServer = await cicsTree.findItem(profileName);
+            expect(wiremockServer).exist;
+            cicsTree.takeScreenshot();
+          });
+
+        it("Should List The Regions Under Wiremock Profile", async () => {
```
davenice, line 90:
```suggestion
        it("Should List The CICSplexes Under Wiremock Profile", async () => {
```

davenice:
Looks good - I made two minor wording suggestions which you could accept through this PR's web interface.

Then do you want to make it non-draft and I'll approve?

davenice:
Also you'll need to do a rebase (Rushabh can help)
And update the description to point at your issue - if you write "Working towards issue #NUMBER" then it won't close the issue, but will associate with it, assuming you've more work to do to complete that issue.

AndrewTwydell (approved):
Looks good! Thanks for this Sneha! 😄

almightyrush (approved):
LGTM

PR #248, opened 2025-03-13
Add resource metadata utilities

AndrewTwydell:
**What It Does**
Add resource metadata utility methods for each resource type. These metadata files follow a strict type to have attributes like Resource name and Human readable name, as well as utility methods for getting default filters and icon values. 

It also adds interfaces for each resource type with the attributes we reference in our code.


davenice (approved):
I think this looks good. Andrew and I talked through some ideas about trying to firm up the contexts because they become API. And about whether the interfaces lock us into particular versions of CICS, which we think they won't.

Issue #250, opened 2025-03-13
[CICS] Lack of logging in the CICS VS Code extension


enamkhan:
There's no real logging in the CICS extension. This makes it hard to be clear on what's happening with customers.

We should work out a way to collect logging (perhaps through the output view like Zowe Explorer) but probably in a different category to Zowe Explorer?

Having some logging of the requests that are made and the response codes would be a really good first step. Also logging the version of the extension somewhere at startup.

davenice:
@enamkhan - do you think we've got enough logging now to close this off?

PR #255, opened 2025-03-19
Added some more logging

enamkhan:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**
(https://github.com/zowe/cics-for-zowe-client/issues/250)

### DIFF
--- a/packages/vsce/src/utils/resourceUtils.ts
+++ b/packages/vsce/src/utils/resourceUtils.ts
@@ -28,6 +28,8 @@ export async function runGetResource({
   cicsPlex?: string;
   params?: { criteria?: string; parameter?: string; queryParams?: IResourceQueryParams };
 }) {
+  CICSLogger.trace("resourceUtils.runGetResource called");
+
   const resourceParams = {
     name: resourceName,
     ...(regionName && { regionName: regionName }),
@@ -38,7 +40,6 @@ export async function runGetResource({
   };
 
   logResourceRequest({
-    call: "runGetResource",
     resourceName,
     cicsPlex,
     regionName,
@@ -61,6 +62,7 @@ export async function runGetResource({
   }
 
   // Making a second attempt as ltpa token has expired
+  CICSLogger.debug("Retrying as validation of the LTPA token failed because the token has expired.");
   session.ISession.tokenValue = null;
   return await getResource(session, resourceParams, requestOptions);
 }
@@ -81,6 +83,8 @@ export async function runPutResource(
   },
   requestBody: any
 ) {
+  CICSLogger.trace("resourceUtils.runPutResource called");
+
   const options: IGetResourceUriOptions = {
     cicsPlex: cicsPlex,
     regionName: regionName,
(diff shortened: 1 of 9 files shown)


davenice (changes requested):
I think it's worth being reasonably careful with the logging that we add and we need some context fairly often - what is the thing we're operating on? What did the user click on?

Otherwise I don't think the logs will turn out to be very helpful.

Given the difficulties customers have around logging in, connections, and schema, I think those would be really interesting areas to focus on. I'm happy to see an example of what a log would like to see can we identify what the user is doing.

e.g. if they're setting a filter, it's probably most useful if we know what filter they're setting on what tree, otherwise they're bound to raise a ticket saying "I set a filter with an @ sign in and it doesn't work" but we won't really be able to identify what's going on.

Is that kind of thing doable or a struggle with this framework?

davenice:
This looks good with some trace with inserts - keen to see what it looks like in practice!

Reflecting on it a bit I wonder if we should do something like have the "called" ones start with a ">" this is what happens in Java logs often and it might lend itself better to syntax highlighting log files.

But really I think I'm mainly keen to see an example and think on it from there!

packages/vsce/src/utils/resourceUtils.ts
```diff
@@ -37,13 +39,11 @@ export async function runGetResource({
...
   };
 
-  logResourceRequest({
-    call: "runGetResource",
-    resourceName,
-    cicsPlex,
-    regionName,
-    ...params
-  });
+  CICSLogger.debug("GET request - Resource [" + resourceName + "]" +
```
AndrewTwydell, line 42: Wondering if specifying the parameters regardless of if they exist would be beneficial or not. 

Option A, `Resource [CICSProgram], Region [MYREG1]`
Option B, `Resource [CICSProgram], CICSPlex [undefined], Region [MYREG1]`

Technically option A shows that the plex name wasn't there, which would be why the request is failing, but option B shows that in a clearer way. 
On the other hand, when you're not wanting a plex name, it could be considered messy to have undefined's thrown around. But again, that's what you expect it to be so maybe that's okay?

packages/vsce/src/utils/resourceUtils.ts
```diff
@@ -61,6 +61,7 @@ export async function runGetResource({
   }
 
   // Making a second attempt as ltpa token has expired
+  CICSLogger.debug("Retrying as validation of the LTPA token failed because the token has expired.");
```
AndrewTwydell, line 64: Wonder if these messages should be constants, seeing as you added that messages file.

```suggestion
  CICSLogger.debug("Validation of the LTPA token failed because the token has expired. Retrying with basic authentication.");
```

AndrewTwydell (commented):
Generally looks fine, few minor comments. 

Also, wonder if there's a way we can turn on some built-in way of printing what methods were called, rather than having to put a trace line in every method? Don't know if this is a thing that exists, but it's looking like we may want the majority of methods traced so 🤷‍♂️

AndrewTwydell (approved):
Looks great! 😄

davenice (approved):

PR #256, opened 2025-03-21
wiremock initial mappings and scenarios

almightyrush:
**What It Does**
Basic mappings and scenarios for disable/ enable program and new copy program.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


davenice (approved):
This looks good to me. 

It will be good to see how sustainable the structure we've got here is! Will we end up with too many xml files and get confused about what goes with what scenario etc.

But I can see that for now we want to reuse the same 'standard' xml responses with several scenarios.

packages/vsce/__tests__/__e2e__/01_EditProjectTeamConfiguration.test.ts
```diff
@@ -88,4 +89,5 @@ describe("Create Project Level Team Configuration File Scenario", () => {
 
     cicsTree.takeScreenshot();
   });
+  resetAllScenarios()
```
AndrewTwydell, line 92: Thinking we should put this in an `after` block, just to be consistent with the test infrastructure setup etc.

```suggestion
  after(() => {
  resetAllScenarios()
  })
```

AndrewTwydell (approved):
Yeah looks good! Assuming the next step is using these mocks in the tests more than we are currently?

PR #269, opened 2025-04-17
Don't require notices for release

davenice:
Now that we're building the notices on PRs and not on release, we don't want to predicate the release on the notices having happened (because they won't have done...!)


AndrewTwydell:
Doh! 🤦🏼 Thanks for sorting that!

PR #274, opened 2025-04-30
Resources visibility can be toggled from VScode settings.

almightyrush:
**What It Does**
CICS resource visibility can be toggled from the VS code settings.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

image


AndrewTwydell:
This looks great and works well! Wondering if we could group the resources more closely, something like this...
Screenshot 2025-05-07 at 3 05 40 PM

packages/vsce/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
 
+## Recent Changes
+
+- Feature: CICS Resources visibilty can be toggled from the VS Code settings [#274](https://github.com/zowe/cics-for-zowe-client/pull/274)
```
AndrewTwydell, line 7: Lets link the issue here, rather than the PR

packages/vsce/package.json
```diff
@@ -417,6 +418,15 @@
         "command": "cics-extension-for-zowe.purgeTask",
         "title": "Purge Task",
         "category": "IBM CICS for Zowe Explorer"
+      },
+      {
+        "command": "cics-extension-for-zowe.toggleResourceSetting",
+        "title": "Click to customize visible CICS Resources",
```
AndrewTwydell, line 424: I don't think we tend to put "Click to..." on each button's tooltip.
```suggestion
        "title": "Toggle visible CICS Resources",
```

packages/vsce/package.json
```diff
@@ -857,6 +872,61 @@
...
+              "description": "CICS URI Map"
+            },
+            "WebService": {
+              "type": "boolean",
+              "description": "CICS Web Service"
+            }
+          },
+          "default": {
+            "Program": true,
+            "LocalTransaction": true,
```
AndrewTwydell, line 920:
```suggestion
            "Transaction": true,
```

AndrewTwydell: I think it needs to be this to match the key above??

AndrewTwydell (changes requested):
Couple suggestions but looking great!

AndrewTwydell:
Also have we thought much about the behaviour of commands like `Inquire Program` on a transaction, or `inquire transaction` on a task, if the target tree is not visible? Currently errors which isn't ideal. 

I see options are:
a) back out of the command if the target tree is not visible `(if !config.get(programs) => infomessage("programs hidden"))`
b) add the target tree in if it's not there before revealing the program
c) see if there's a way of disabling the command in the package.json (`when: config.get(zowe resources.programs) is true`)

davenice:
Great! A few questions/thoughts...
- Can we show TCP/IP Services with the `/` in the GUI toggle?
- I don't think we need to say `Enable <resource>` on the toggle, because we've explained at the top of the settings what we're doing. However it doesn't look like the code will do the same as your screenshot, so maybe you've already tweaked that!
- Does this also hide the "All <resources>" options at the plex level? I think ideally it would.

Looks neat!

almightyrush:
> Great! A few questions/thoughts...
> 
> * Can we show TCP/IP Services with the `/` in the GUI toggle?
> * I don't think we need to say `Enable <resource>` on the toggle, because we've explained at the top of the settings what we're doing. However it doesn't look like the code will do the same as your screenshot, so maybe you've already tweaked that!
> * Does this also hide the "All " options at the plex level? I think ideally it would.
> 
> Looks neat!

- changed to TCP/IP
- The screenshot is the older one, I will upload the updated :-)
- Yes it will hide the All Resource type at plex level

Thanks !

AndrewTwydell (approved):
Looks great - can you make that 1 change and then we're good to go!

almightyrush:
https://github.com/user-attachments/assets/6f12490e-dd0e-45e8-b0b6-da9ef87c0c3e

This is how it will look like!!

packages/vsce/src/trees/CICSRegionTree.ts
```diff
@@ -65,17 +54,20 @@ export class CICSRegionTree extends TreeItem {
...
-        new CICSTransactionTree(this),
-        new CICSLocalFileTree(this),
-        new CICSTaskTree(this),
-        new CICSLibraryTree(this),
-        new CICSPipelineTree(this),
-        new CICSTCPIPServiceTree(this),
-        new CICSURIMapTree(this),
-        new CICSWebServiceTree(this),
-      ];
+        config.get<boolean>("Program", true) ? new CICSProgramTree(this) : null,
```
enamkhan, line 61: Not major, but you could also do `config.get<boolean>("Program", true) && new CICSProgramTree(this)` so that the else clause is no longer needed.

enamkhan (approved):
LGTM!

PR #275, opened 2025-05-01
Axios: Remove it by adding a few enhancements :yum:

zFernand0:
**What It Does**

This is mostly an easy way to remove it, but it adds a bunch of enhancements on the V2 LTS branch

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] added/updated automated tests
- [ ] updated the changelog
- [ ] considered whether the docs need updating
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #276, opened 2025-05-01
Axios: Remove it without adding enhancements :yum:

zFernand0:
**What It Does**

I believe this is the preferred way of removing axios from the V2 LRS branch since this does not add any enhancements 🙏  

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

I will need some help testing these changes since I'm still recovering my environments from a few other destructive tests I was doing.

### DIFF
--- a/packages/vsce/src/utils/profileManagement.ts
+++ b/packages/vsce/src/utils/profileManagement.ts
@@ -9,11 +9,11 @@
  *
  */
 
-import { ZoweExplorerApi, ZoweVsCodeExtension, imperative } from "@zowe/zowe-explorer-api";
+import { ZoweVsCodeExtension, imperative } from "@zowe/zowe-explorer-api";
 import { window } from "vscode";
 import cicsProfileMeta from "./profileDefinition";
 import { CICSPlexTree } from "../trees/CICSPlexTree";
-import { getResource } from "@zowe/cics-for-zowe-sdk";
+import { getCache, getResource } from "@zowe/cics-for-zowe-sdk";
 
 export class ProfileManagement {
   private static zoweExplorerAPI = ZoweVsCodeExtension.getZoweExplorerApi();
@@ -41,7 +41,7 @@ export class ProfileManagement {
   }
 
   public static async profilesCacheRefresh() {
-    const apiRegiser: ZoweExplorerApi.IApiRegisterClient = ProfileManagement.getExplorerApis();
+    const apiRegiser = ProfileManagement.getExplorerApis();
     await ProfileManagement.getProfilesCache().refresh(apiRegiser);
   }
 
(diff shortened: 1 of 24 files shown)


davenice:
Thanks Fernando - we can give this a test.

For our background, is there a pressing need or imperative to remove Axios? I think updating the LTS client is probably low risk because most users will be on the latest, and we can test it fairly well, but just for our understanding.

👍

zFernand0:
> For our background, is there a pressing need or imperative to remove Axios? I think updating the LTS client is probably low risk because most users will be on the latest, and we can test it fairly well, but just for our understanding.

Hey Dave! 🥳 
No, there isn't a pressing need to remove it. It's mostly following the same practices as with the latest release. 😋 
I opened 2 PRs, but feel free to close the other one (#275) as it adds some enhancements 😅

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -69,185 +66,158 @@ export class ProfileManagement {
...
     const infoLoaded: InfoLoaded[] = [];
-
-    https.globalAgent.options.rejectUnauthorized = profile.profile.rejectUnauthorized;
+    const session = ProfileManagement.getSessionFromProfile(profile.profile);
 
     if (profile.profile.cicsPlex) {
+      const managedRegionsJson = await getResource(session, {
+        resource: "CICSManagedRegion",
+        plex: profile.profile.cicsPlex,
+        region: profile.profile.regionName,
```
AndrewTwydell, line 95: The `IResourceParms` interface asks for these attributes which will be for each getResource method call.

```suggestion
        name: "CICSManagedRegion",
        cicsPlex: profile.profile.cicsPlex,
        regionName: profile.profile.regionName,
```

packages/vsce/src/utils/profileManagement.ts
```diff
@@ -69,185 +66,158 @@ export class ProfileManagement {
...
-          try {
-            const singleRegion = await ProfileManagement.makeRequest(`/CICSRegion`, config);
-            const jsonFromXml = JSON.parse(xml2json(singleRegion.data, { compact: true, spaces: 4 }));
-            if (!jsonFromXml) {
-              throw error;
-            }
-            const returnedRegion = jsonFromXml.response.records.cicsregion._attributes;
+        } else {
+          // Not Plex
+          const singleRegionJson = await getResource(session, {
```
AndrewTwydell, line 193: The implementation of `getResource` at this version requires a region name causing failures.
```
ImperativeExpect.toBeDefinedAndNonBlank(parms.regionName, "CICS Region name", "CICS region name is required");
```

Being unfamiliar with maintaining this V2 version, what's the best approach here? The reason we initially adapted getResource in V3 was to accommodate not specifying a region name 🤔

AndrewTwydell (commented):
@zFernand0 thanks for looking at this. I have a high level question around getResource below that perhaps you can share some insight into the best approach.

packages/vsce/src/utils/profileManagement.ts
AndrewTwydell: Ahhh the second PR adds these changes in..! Perhaps that's necessary to complete the removal of Axios

zFernand0: Hey @AndrewTwydell,

This PR is meant to just remove axios without introducing enhancements. 

The other PR (#275) should not be merged as it contains enhancements for the V2 (maintenance) branch 😅 
I opened it just as an example, but would love to close it in favor of this one 😅

packages/vsce/src/utils/profileManagement.ts
zFernand0: I'm hoping to avoid changes to these interfaces (if possible) since this is the V2 (maintenance) release stream 😓

AndrewTwydell: I don't think this is _changing_ any interfaces - the `IResourceParms` interface that `getResource` asks for as a paramter has `name`, `cicsPlex` and `regionName` so presumably that's the shape we'll need to pass to use `getResource` instead of Axios. 

We're also searching here for regions within a plex, where we don't necessarily have a regionName to pass (`profile.profile.regionName` could be undefined).

The V2 implementation of `getResource` requires a regionName to be provided - the rational behind removing that required argument in V3 was so we could make requests like this; to get region / plex resources.

I'm not sure what the best solution is here. To use `getResource` everywhere in this file rather than axios, we'd need to allow `getResource` in V2 to NOT require a regionName, like V3. Making a currently mandatory field optional wouldn't be considered a breaking change I assume? As those using this version can still use it with no changes.

zFernand0: Resolved in e46c3f2b0ec33b8a6f6b0a19db7690d565ae77af

davenice (approved):
I think this looks reasonable - I've tested all the clicking around I can think of. It's quite a lot of change which makes it hard to be sure there aren't bugs, but I don't think that many people will be using v2.

PR #286, opened 2025-05-06
fix: autostore credentials

AndrewTwydell:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


davenice (approved):
Looks reasonable to me.

I've tested it from a dev environment in debug mode; I had some connections where no attributes were set secure and autostore was false.

I got prompted for the user/pass and then my session was established and I could click around, all seemed well.

I hit refresh on the CICS tree and got prompted again.

As discussed, we still get a 401 error once if you haven't got a saved password, but this is a good step in the right direction.

THANKS!

PR #288, opened 2025-05-13
Use new Zowe extender API to trigger create profile, add schema version

davenice:
**What It Does**
Updates to use the Zowe extender API provided with Zowe Explorer 3.2.0
No discernable change .... so can we avoid change-log and issue? 🤔 
Our end 2 end tests will cover the testing very well!

Also adding a version into the profile schema - https://github.com/zowe/cics-for-zowe-client/issues/291 - which should have a change log.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/cli/NOTICE
+++ b/packages/cli/NOTICE
@@ -1446,6 +1446,29 @@ ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 ------
 
+** inherits; version 2.0.4 -- https://github.com/isaacs/inherits#readme
+Copyright (c) Isaac Z. Schlueter
+
+The ISC License
+
+Copyright (c) Isaac Z. Schlueter
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+PERFORMANCE OF THIS SOFTWARE.
+
+
+
+------
+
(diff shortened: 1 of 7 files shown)


davenice:
@JillieBeanSim - I think this is what you were intending. Is there any testing you'd suggest for the version addition? It seems like the version doesn't do anything unless you change it. And we would change it if we updated the schema.

davenice:
I'll need to update the snapshot in the tests if we agree I'm doing the right thing.

AndrewTwydell:
Got this message - I think you've mentioned the string templating before and is out of our control right??

Screenshot 2025-05-14 at 11 05 22 AM

davenice:
> Got this message - I think you've mentioned the string templating before and is out of our control right??
> 
That will be fixed by this Zowe Explorer PR - https://github.com/zowe/zowe-explorer-vscode/pull/3648

davenice:
This has got in a pickle and lost all the checks. I dislike doing this, but I'm going to create a new PR.

AndrewTwydell (approved):

JillieBeanSim (commented):
Hey @davenice the version update works great, I see the update in /.zowe/extenders.json file with version
I do see that the create new team config (I did project level for testing) only had the core Zowe profile types and no registered extenders including CICS profile.  Are you seeing these extender types being added?

davenice:
> Hey @davenice the version update works great, I see the update in /.zowe/extenders.json file with version I do see that the create new team config (I did project level for testing) only had the core Zowe profile types and no registered extenders including CICS profile. Are you seeing these extender types being added?

Ugh. You're right - I hadn't really registered that the CICS stanza got added normally so didn't spot that it wasn't now.

That's a shame - shall I raise something against ZE?

davenice:
@Snehabayya do you think you could extend our e2e tests that create profiles, so that they verify that the word `cics` has been added to the profile after creating? I hadn't realised this was part of the process.

If you're deep in other stuff then I'll have a go :-D

AndrewTwydell:
You're right @JillieBeanSim! Doing some digging, it seems [this block](https://github.com/zowe/zowe-explorer-vscode/blob/1d5e95b04acdc084aa7b87277710a1b4dc57b811/packages/zowe-explorer-api/src/vscode/ZoweVsCodeExtension.ts#L384) is getting the profile types to add to the schema. 

`ZoweVsCodeExtension.profilesCache.getCoreProfileTypes()` obviously returns the zosmf, ssh, and tso schemas.

`ZoweVsCodeExtension.profilesCache.getConfigArray()` looks like you're expecting it to return the extenders' schemas, but that is returning nothing for us. 

Either the new method we're calling isn't working quite right, or the way we're registering ourselves is not how you'd expect / the correct way of doing it.

Snehabayya:
> @Snehabayya do you think you could extend our e2e tests that create profiles, so that they verify that the word `cics` has been added to the profile after creating? I hadn't realised this was part of the process.
> 
> If you're deep in other stuff then I'll have a go :-D

Hey Dave, in the test scenario where I'm adding the WireMock profile, it already includes `cics` since the profile is defined earlier. Would you like me to add an expect statement to check if the profile contains the word `cics` before writing it to the configuration file?

davenice:
I've split the schema work off into https://github.com/zowe/cics-for-zowe-client/pull/323

I'll raise a new PR for the extender API once I've got it working correctly.

PR #293, opened 2025-05-15
Run notice updates on release

AndrewTwydell:
**What It Does**
Runs the NOTICE file updates as part of the release, pushing to the main branch after the PR is merged - the same behaviour as dep version and changelog updates.

I ran the steps as part of another stage to test the generation and commiting worked as expected - it did. 
There is no way to 100% guarantee this runs as part of the release but I am confident. We will watch carefully on the next release to ensure it works okay.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,14 +2,7 @@ name: Zowe CICS Main CI
 
 on:
   push:
-    paths:
-      - .github/**
-      - packages/**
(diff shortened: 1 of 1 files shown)


davenice (approved):
What is our plan around testing?

Do we need to make sure we come back after the next release and double check it's worked?

enamkhan (approved):
LGTM!

PR #294, opened 2025-05-16
Added an api call that returns the supported resources config

enamkhan:
Added an api call for getting the supported resources, this does not change any functionality. 

**How to Test**
Have tested by calling this method.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/extension.ts
+++ b/packages/vsce/src/extension.ts
@@ -19,6 +19,7 @@ import { getZoweExplorerVersion } from "./utils/workspaceUtils";
 import { getCommands } from "./commands";
 import { CICSLogger } from "./utils/CICSLogger";
 import { CICSMessages } from "./constants/CICS.messages";
+import { CicsCmciConstants } from "@zowe/cics-for-zowe-sdk";
 
 /**
  * Initializes the extension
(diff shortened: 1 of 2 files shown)


packages/vsce/src/extension.ts
```diff
@@ -177,6 +178,22 @@ export async function activate(context: ExtensionContext) {
...
+    api: {
+      configuration: {
+        supportedResources: [
+          CicsCmciConstants.CICS_CMCI_LOCAL_FILE,
+          CicsCmciConstants.CICS_PROGRAM_RESOURCE,
+          CicsCmciConstants.CICS_CMCI_LOCAL_TRANSACTION,
+          CicsCmciConstants.CICS_TCPIPSERVICE_RESOURCE,
+          CicsCmciConstants.CICS_LIBRARY_RESOURCE,
+          CicsCmciConstants.CICS_URIMAP,
+          CicsCmciConstants.CICS_CMCI_TASK,
```
AndrewTwydell, line 192: Is there a reason we've only got a subset of the reasources in the tree, or should we add the rest? (7 out of 9)

Screenshot 2025-05-16 at 12 37 51 PM

AndrewTwydell (commented):
How hard would it be to have a unit test to assert the return value of the `activate` method? Not only will this be good practice, but will prevent us _changing_ the API that extenders may rely on, causing breaking changes. If we have a test for this, we can't change it which will be good!

PR #308, opened 2025-06-06
Update E2E tests for program enable/disable, new copy and attribute display actions

Snehabayya:
**What It Does**
Performs disable/enable program, new copy and show attributes actions on programs in region IYCWENK1 in plex CICSEX61

**How to Test**
npm run test:e2e

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
+++ b/packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
@@ -22,6 +22,7 @@ import {
 import {
   clickCollapseAllsIconInCicsTree,
   closeAllEditorsTabs,
+  expectTreeItemIsSelected,
   getCicsSection,
   getPlexChildIndex,
   getPlexChildren,
@@ -116,6 +117,7 @@ describe("Test Suite For Performing Actions On The Programs In CICSEX61", () => 
     it("Should Disable The Program C128N", async () => {
       // Navigate to the C128N program in the tree
       await sendArrowDownKeyAndPressEnter(5);
+      await expectTreeItemIsSelected(C128NProgram);
 
       // Run the disable command from the command palette
       // And get the programs in region IYCWENK1 in plex CICSEX61 to get updated state
@@ -135,6 +137,7 @@ describe("Test Suite For Performing Actions On The Programs In CICSEX61", () => 
 
     it("Should Enable The Program C128N", async () => {
       await sendArrowDownKeyAndPressEnter(5);
+      await expectTreeItemIsSelected(C128NProgram);
 
       // Run the enable command from the command palette
       // And get the programs in region IYCWENK1 in plex CICSEX61 to get updated state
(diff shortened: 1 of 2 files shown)


AndrewTwydell (approved):
Looks awesome, thanks for adding those! 

I think we're going to run into some issues when we remove a lot of the commands from the command pallette as you're currently relying on them being there. I think we will hide them based on an environment variable or something, but we'll handle that when we come to it 😄

packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    it("Should Check If The Program C128N Is Present In Region IYCWENK1", async () => {
+      C128NProgram = await findProgramByLabel(programs, C128N);
+      expect(C128NProgram).not.undefined;
+      expect(await C128NProgram?.getLabel()).contains(C128N);
+      cicsTree.takeScreenshot();
+    });
+
+    it("Should Disable The Program C128N", async () => {
+      // Navigate to the C128N program in the tree
+      await sendArrowDownKeyAndPressEnter(5);
```
davenice, line 118: this is very likely to break in the future if we change what's in the tree! is it essential this is C128N? if so it would be worth asserting that.

davenice (commented):
Minor suggestion! Some nice refactoring / methods here 👍

packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
Snehabayya: The program label was checked in the previous test case. Here I'm using `sendArrowDownKeyAndPressEnter()` to select the program, since `click()` and `select()` are not functioning as expected.

packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
```diff
@@ -0,0 +1,249 @@
...
+  let cicsex61Children: TreeItem[];
+  let regionsIndex: number;
+  let regions: TreeItem[];
+  let regionK1Resources: TreeItem[];
+  let programs: TreeItem[];
+  let regionIYCWENK1Index: number;
+  let programsResourceIndex: number;
+
+  before(async () => {
+    await sleep(2000);
```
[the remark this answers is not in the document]
Snehabayya: Yeah, these are required initially to give VS Code some time to finish setting up

almightyrush (approved):
LGTM

packages/vsce/__tests__/__e2e__/03_PerformActionsOnPrograms.test.ts
davenice: hey - i understand that. Here's the scenario I'm thinking of:

Somebody else alters the mocks to add a new program - say "ANEXMPL".
Now, you pressing down 5 times will not take you to C128N, it'll take you to the program before this.

Your test will carry on and will disable whatever the program is before C128N.
Then your test will check that C128N is disabled, and it will fail saying that we expected C128N to be disabled.

Then somebody will have to debug the test wondering why disable has stopped working.

However the real problem will be that the cursor didn't end up on C128N when it should have done - if your test code was to move down 5 rows _and then check whether you were on C128N_ then i think that validates the assumption you're making (that C128N is the 6th program) before you start attempting to disable

PR #311, opened 2025-06-09
Add bundle and bundleparts to tree

AndrewTwydell:
**What It Does**
Adds bundle and bundlepart resources to the tree, with enable and disable actions.

Requires new icons!

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/vsce/package.json
+++ b/packages/vsce/package.json
@@ -393,6 +393,14 @@
           "when": "config.zowe.cics.showAllCommandsInPalette"
         },
         {
+          "command": "cics-extension-for-zowe.enableBundle",
+          "when": "config.zowe.cics.showAllCommandsInPalette"
+        },
+        {
+          "command": "cics-extension-for-zowe.disableBundle",
+          "when": "config.zowe.cics.showAllCommandsInPalette"
+        },
+        {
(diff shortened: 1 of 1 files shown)


davenice:
I notice that in all bundles something odd happens - it appears to be if there's a duplicate across two regions:
image

The bundleparts seem to get confused. I'm guessing we don't make quite the right query but I haven't debugged it yet.

There a slight annoyance which happens with every resource type where when you expand it says "2 of 2" rather than showing the region that it's from. But that's a pre-existing bug/niggle so let's not address that right now.

almightyrush (approved):
Lets get this bundle in !!!

davenice (approved):

davenice:
> I notice that in all bundles something odd happens - it appears to be if there's a duplicate across two regions: image
> 
> The bundleparts seem to get confused. I'm guessing we don't make quite the right query - we're not filtering correct on the region name of the parent - but I haven't debugged it yet. It doesn't seem to affect Library in quite the same way.
> 
> There a slight annoyance which happens with every resource type where when you expand it says "2 of 2" rather than showing the region that it's from. But that's a pre-existing bug/niggle so let's not address that right now.

Andrew put in a fix for this 👍

PR #313, opened 2025-06-09
Sticky Header Implementation in Resource Inspector View

chinmdas:
**What It Does**
This PR brings in a new UI feature where Resource Inspector now has sticky headers. This makes the look and feel more elegant.

Screen Recording 2025-06-09 at 5 40 13 PM

### DIFF
--- a/packages/vsce/src/webviews/css/style.css
+++ b/packages/vsce/src/webviews/css/style.css
@@ -5,14 +5,40 @@
   gap: 1rem;
 }
 
+.table-header1 {
+  position: fixed;
+  top: 0;
+  padding-left: 0.625rem;
+  padding-right: 4.188rem;
+  left: 2.25rem;
+  right: 6.25rem;
+}
+
+.padding-top-20 {
+   padding-top: 1.25rem;
+}
+
 .line {
   line-height: 0.5;
   font-size: small;
 }
 
 .header-cell-1 {
-  padding-left: 10px;
   vertical-align: middle;
+}
+
+.thead-2{
+  position: sticky;
+}
+
+.th-2 {
+  width: 100% !important;
+  padding-left: 1.063rem !important;
+}
+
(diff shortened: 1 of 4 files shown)


packages/vsce/src/webviews/css/style.css
```diff
@@ -5,39 +5,155 @@
...
+
+.vertical-align-sub {
+  vertical-align: sub;
+}
+
+.border-collapse {
+  border-collapse: collapse;
+}
+
+body.vscode-light table,
```
almightyrush, line 104: Most of the css properties are repeated for light and dark theme.
It can be combined to reduce the redundancy. 
Something like this -->
```suggestion
/* Combined common table styling for both themes */
body.vscode-light table,
body.vscode-light th,
body.vscode-light tr,
body.vscode-light td,
body.vscode-dark table,
body.vscode-dark th,
body.vscode-dark tr,
body.vscode-dark td {
  padding: 0.5rem;
  font-family: var(--vscode-font-family);
  color: var(--vscode-foreground);
}

/* Light theme specific */
body.vscode-light th,
body.vscode-light thead {
  background-color: #f0f0f0;
  text-align: left;
  color: var(--vscode-foreground);
}
body.vscode-light th {
  width: 73.75rem;
}
body.vscode-light tr:nth-child(even) {
  background-color: #f2f2f2;
}

/* Dark theme specific */
body.vscode-dark th,
body.vscode-dark thead {
  background-color: rgb(40,40,40);
  text-align: left;
  color: var(--vscode-foreground);
}
body.vscode-dark th {
  width: 73.75rem;
}
body.vscode-dark tr:nth-child(even) {
  background-color: rgba(204, 204, 204, 0.04);
}
body.vscode-dark input {
  background-color: rgb(60, 60, 60);
}

```

AndrewTwydell (approved):

Issue #283, opened 2025-05-06
[CICS] Introduce “Default Region” status bar widget and quick pick framework


davenice:
**Describe your enhancement idea**

Reducing the amount of scrolling and finding in the CICS tree will improve a user's experience. During a development session a user is generally focussed on a particular region that they are preparing to test in, or develop against, before rolling out to a wider set of regions at a later point.

Introduce a status bar widget that references a default region the user is focussed on. This would allow VS Code-native command palette interactions with low friction - for instance a command for "Show logs for region" where the user is offered the option to use quick picks to choose the region they're interested in, but the default region is pre-selected.

The status bar icon might look something like this:
<img width="395" alt="Image" src="https://github.com/user-attachments/assets/270fd94b-68e8-4ec8-bf70-b37b67a32cca" />

When the user clicks the status bar, they are prompted via quick picks to choose a profile, plex, and region to use as their default.

This flow can then be used for any command that needs a region. The simplest example is show logs for region.

The flow here would be:
- User chooses show logs from region from the command palette
- Quick pick offers the user the default region or the option to pick a different region
- If the user chooses the default region, the logs for the default region are shown
- If the user chooses the option to pick a different region, the quick pick for profile/plex/region is run through

We need to consider what will happen if the default region's profile is not connected and how we trigger that to happen - how do we indicate that it wasn't possible?
Or if the user changes their set of profiles and the profile the default region comes from is no longer available.

Indicating the default region in the tree is something @Rajat-Sahrawat has been considering.

Storing the default region information in the preferences would sow the seeds for it to be portable/shareable between users working on an application, if they chose.

**Delivery details**
The delivery of this could be staged something like this:
- A "show region logs" command that shows a set of quick picks allowing the user to choose profile/plex/region
- Refactoring the quick picks out to a separate "set default region" command
- Adding in a status bar that shows the default region
... and probably more steps!

PR #318, opened 2025-06-10
select active region using quickpick

almightyrush:
**What It Does**
select Active region using quick pick and return profile, plex, session and region name

**How to Test**
open command palette 
click on `Select Active CICS region`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


almightyrush:
> Hey @almightyrush - do you want to mark this as "working towards" #283 and also i think let's put it under the same feature flag as the resource inspector, given that we're not change-logging it?
> 
> Is it a quick job to add a couple of e2e tests for it, or is that a long job?

Yes, I will link the issue with this PR. 
If you want e2e to go with this PR, it may take some time. I can do that in separate PR, if thats okay.

almightyrush:
Also for the next bit I will be working on the session handling, will update the mural on what ideas I have to propose

packages/vsce/package.json
```diff
@@ -63,6 +63,11 @@
       ]
     },
     "commands": [
+      {
+        "command": "cics-extension-for-zowe.selectFocusRegion",
+        "title": "Select Focus CICS Region",
```
davenice, line 68: i think this should be Update - because select could imply "pick it in the tree"

almightyrush: Can it be Select/Update .. 
As to consider for the very first time selecting the region

davenice: ok - let's call it Set?

Select/Update is very clunky! I don't think people will generally use the command to set it, but it's there.

packages/vsce/package.json
AndrewTwydell, line 1: Lets hide this command from the command pallete unless the resourceInspector config is set

AndrewTwydell:
```json
"commandPalette": [
  {
    "command": "cics-extension-for-zowe.selectFocusRegion",
    "when": "config.zowe.cics.resourceInspector"
  }
],
```

AndrewTwydell (approved):
Looks good, thanks @almightyrush !

PR #319, opened 2025-06-10
Fix quotes in criteria for library dataset

Khushboo-Sharma-110597:
https://github.com/zowe/cics-for-zowe-client/issues/320


davenice:
Great, thanks @Khushboo-Sharma-110597 - let's have a look at this tomorrow. Please can you raise a bug that explains the issue and link it with this PR. I'll improve the description of the PR because otherwise it's difficult to explain things at the Zowe scrum :-D

davenice:
Also if this is something customers could hit in the field, we should add a change-log entry to the vsce package.

davenice:
@Khushboo-Sharma-110597 I don't think we need a change-log because we haven't shipped the code with this bug in yet. 🎉

davenice (approved):
This fixes something that I don't believe users will come across in the currently shipped version, so I don't think we need a changelog.

AndrewTwydell (approved):
Nice! Thanks 😄

PR #327, opened 2025-06-12
Do not show in the command palette commands that require a specific selection to activate

harshithajahagirdar:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


almightyrush (changes requested):
You have done very well. Based on the flag set from the settings.json to hide the least priority commands from the palette.
I think that creating new command is not required, the extra set of code can be minimised.
What I would do is, go to `../sdk/ICICSExtenderConfig.ts` file 
and insert this 
```
showAllCommand: {
      enabled: boolean;
};
```

Next I will update it in the `../vsce/CICSExtenderApiConfig.ts `as an implementation of `ICICSExtenderConfig`
```
showAllCommand: {
      enabled: false
}
```

Just update the when condition in the commandPalette to: 
from "`config.cics-extension-for-zowe.showAllCommandsInPalette`" to `config.zowe.cics.showAllCommandsInPalette`. 

Thats it all done!!.
It can now be toggled from the `settings.json.`

Let me know what think about it. :-)

packages/vsce/__tests__/__unit__/commands/showAllCommandsInPalette_true.unit.test.ts
AndrewTwydell, line 1: I'm not sure what this is actually testing. You're setting values on a mocked method and asserting they are what you set. Don't think this is needed.

packages/vsce/__tests__/__unit__/commands/showAllCommandsInPalette_false.unit.test.ts
AndrewTwydell, line 1: I'm not sure what this is actually testing. You're setting values on a mocked method and asserting they are what you set. Don't think this is needed.

packages/vsce/__tests__/__e2e__/util/globalMocks.ts
```diff
@@ -85,3 +96,15 @@ export async function sendArrowDownKeyAndPressEnter(times: number): Promise<void
   }
   await driver.actions().sendKeys(Key.ENTER).perform();
 }
+
+export async function openCommandPaletteAndRun(command: string) {
```
AndrewTwydell, line 100: I really like this function, seems useful from lots of places! 😄 

It's a shame the sleeps are necessary - have you tested without them?

packages/vsce/__tests__/__e2e__/20_zowe.cics.showAllCommandsInPalette_true.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  let settings: any = {};
+  try {
+    settings = JSON.parse(await editor.getText());
+  } catch {
+    settings = {};
+  }
+  settings["zowe.cics.showAllCommandsInPalette"] = value;
+  await editor.setText(JSON.stringify(settings, null, 2));
+  await editor.save();
+}
```
AndrewTwydell, line 26: These methods look to be the same elsewhere. Perhaps they can be moved to a utility file somewhere, rather than duplicating?

AndrewTwydell (changes requested):
This looks great! I've made a few notes, but he summary is...

- I think there are a few files/changes that we don't need, like the extender config files.

- The name of the setting should be consistent...
  - In `settings.json` you've got `cics-extension-for-zowe.showAllCommandsInPalette`
  - In the test files, you've got `zowe.cics.showAllCommandsInPalette`
  - In `package.json` you've got `zowe.cics.showAllCommandsInPalette`

I think these need to be consistent right? Package.json needs to be `config.` + whatever the setting is called but the rest need to be the same I think?

- There are a lot of test files, some of which seem to be doing the same, or very similar, thing. Perhaps they can be condensed into 1 or 2 files?

packages/sdk/src/doc/ICICSExtenderConfig.ts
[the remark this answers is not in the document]
almightyrush: Aah, it was my misunderstanding. Adding property can be directly used from `package.json`. 
@harshithajahagirdar Remove the references from the extenders

packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
```diff
@@ -44,6 +44,15 @@
         "host": "localhost",
         "rejectUnauthorized": true
       }
+    },
```
almightyrush, line 47: This is not required, the previous tests is already adding the profile and it gets removed in the last test execution.

packages/vsce/package.json
```diff
@@ -260,6 +260,113 @@
...
+        },
+        {
+          "command": "cics-extension-for-zowe.filterResources",
+          "when": "config.zowe.cics.showAllCommandsInPalette"
+        },
+        {
+          "command": "cics-extension-for-zowe.showResourceAttributes",
+          "when": "config.zowe.cics.showAllCommandsInPalette"
+        },
         {
```
almightyrush, line 370: The command `"cics-extension-for-zowe.inspectResource"` is repeated. Its present in the top too.

packages/vsce/__tests__/__e2e__/15_zowe.cics.showAllCommandsInPalette_false.test.ts
almightyrush, line 1: Why there are 2 different files for same test ?
`15_zowe.cics.showAllCommandsInPalette_false.test.ts` and  `23_zowe.cics.showAllCommandsInPalette_false.ts`.
And whats with numbering the files, 12,20,23,30 35. 
Can you please make it in sequence, and combine the files with same test, if serves same purpose

packages/vsce/Untitled-1.js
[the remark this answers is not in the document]
harshithajahagirdar: Yes, I will delete the file. As there were many changes in package.json I created that file to cross check.

packages/vsce/__tests__/__e2e__/15_zowe.cics.showAllCommandsInPalette_false.test.ts
harshithajahagirdar: The thing is I have written 2 scenarios. One where the environment variable is set to true or false, then counting the number of commands. Then will check if it has more than 5 commands in case of true and less than 5 commands in case of false. 
The other scenario is, I am checking for a particular command to be present when its true, but not when its false. Dave suggested to add 2nd scenario just to confirm, things wont break that easily.
I have written different test cases for true and false, as toggling them in the same test case might not give right result all the time. I have tried saving the file, but sometimes it did not reflect as anticipated.

packages/vsce/package.json
harshithajahagirdar: Resolved.

packages/vsce/__tests__/__e2e__/20_zowe.cics.showAllCommandsInPalette_true.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    console.log("Setting it to True");
+    await writeShowAllCommandsInPaletteValue(editor1, true);
+    await sleep(500);
+    await editor1.save();
+    //await sleep(50000);
+    const countTrue = await countCicsCommands();
+    console.log("Commands when setting is True:", countTrue);
+
+    await sleep(500);
+    expect(countTrue).greaterThan(5);
```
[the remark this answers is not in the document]
harshithajahagirdar: I tried doing that, but VS code displays few other commands, even when I am searching with a keyword. And the number of commands to be displayed when the environment variable is set to false might vary in the future(we would want to add or remove some commands).Hence I have written my scenario to check if it has more or less commands.
Screenshot 2025-06-28 at 2 17 56 PM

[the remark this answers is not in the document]
AndrewTwydell: Somewhere else in your testing you got the commands, and fetched their labels right? Could we filter the command labels that are in the command palette to count the number of labels with “IBM CICS for Zowe Explorer”? I understand your issue though

packages/vsce/__tests__/__e2e__/util/globalMocks.ts
harshithajahagirdar: Yes, I have tried. Without sleep, Sometimes it wont reflect the changes.

packages/sdk/src/doc/ICICSExtenderConfig.ts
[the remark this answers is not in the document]
harshithajahagirdar: Removed

packages/vsce/__tests__/__e2e__/resources/test/config-files/zowe.config.json
harshithajahagirdar: Removed.

packages/vsce/__tests__/__unit__/commands/showAllCommandsInPalette_false.unit.test.ts
harshithajahagirdar: I have written a unit test to mock zowe.cics.showAllCommandsInPalette, when the value is false. If you feel its not needed I can remove. Please let me know if you have any suggestions.

packages/vsce/__tests__/__e2e__/20_zowe.cics.showAllCommandsInPalette_true.test.ts
harshithajahagirdar: Added all the functions to globalMocks.ts.

packages/vsce/__tests__/__unit__/commands/showAllCommandsInPalette_false.unit.test.ts
AndrewTwydell: So my logic is... If you changed any of the extension code, renamed the setting, removed it, changed any behaviour, this test will always pass. It's only actually testing methods that are defined in this file. It doesn't test the behaviour of the setting or command palette.

packages/vsce/__tests__/__e2e__/20_zowe.cics.showAllCommandsInPalette_true.test.ts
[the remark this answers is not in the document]
harshithajahagirdar: The function countCicsCommands does the same thing... It opens the command palette, then filters the input by entering “IBM CICS for Zowe Explorer” to the input box. Then it counts the number of commands displayed.

packages/vsce/__tests__/__e2e__/15_zowe.cics.showAllCommandsInPalette_false.test.ts
harshithajahagirdar: okay, I will add both the scenario to the same test case. And I have changes the file name.

PR #334, opened 2025-06-18
Adding sessionHandlers for TreeItem and Focus Region

almightyrush:
**What It Does**
Creates and manages the CICSSessions across the extension. It will reuse the session in between the Focus Region and CICS tree

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/resources/SessionHandler.ts
+++ b/packages/vsce/src/resources/SessionHandler.ts
@@ -17,12 +17,14 @@ export class SessionHandler implements ISessionHandler {
   private sessions: Map<String, CICSSession>;
   private static instance: SessionHandler;
 
-  private constructor() {}
+  private constructor() {
+    this.sessions = new Map<String, CICSSession>();
+  }
+
   // Creating a singleton instance of SessionHandler
   public static getInstance(): SessionHandler {
     if (!SessionHandler.instance) {
       SessionHandler.instance = new SessionHandler();
-      SessionHandler.instance.sessions = new Map<String, CICSSession>();
     }
     return SessionHandler.instance;
   }
--- a/packages/vsce/tsup.config.json
+++ b/packages/vsce/tsup.config.json
@@ -1,11 +1,15 @@
 {
-  "entry": ["src/extension.ts"],
+  "entry": [
+    "src/extension.ts"
+  ],
   "splitting": false,
   "sourcemap": true,
   "clean": true,
   "format": "cjs",
   "dts": true,
-  "external": ["vscode"],
-  "minify": false,
+  "external": [
+    "vscode"
+  ],
+  "minify": "terser",
   "noExternal": []
 }


packages/vsce/src/resources/SessionHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { IProfile, IProfileLoaded } from "@zowe/imperative";
+import { imperative } from "@zowe/zowe-explorer-api";
+import { CICSSession } from "./CICSSession";
+import { ISessionHandler } from "./ISessionHandler";
+
+export class SessionHandler implements ISessionHandler {
+  public sessions: Map<String, CICSSession>;
+  session: CICSSession | undefined;
```
AndrewTwydell, line 19: I'm not sure we need a global `sessions` and `session` object? Think we can remove `session` and just store them all in the map. 

Also, we could make `sessions` private as we have a getter method 😄

packages/vsce/src/resources/SessionHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+  profile: imperative.IProfile;
```
AndrewTwydell, line 20: Same here, we don't need `profile` do we??

packages/vsce/src/resources/SessionHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+  private static instance: SessionHandler;
+
+  private constructor() {}
+  // Creating a singleton instance of SessionHandler
+  public static getInstance(): SessionHandler {
+    if (!SessionHandler.instance) {
+      SessionHandler.instance = new SessionHandler();
+      SessionHandler.instance.sessions = new Map<String, CICSSession>();
```
AndrewTwydell, line 28: I think we could set the initial value of `sessions` in the constructor?

packages/vsce/src/resources/SessionHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return SessionHandler.instance;
+  }
+
+  createSession(profile: IProfile, profileName: string): void {
```
AndrewTwydell, line 33: Is the idea that this is only called from inside this class, from `getSession` if it doesn't exist?? If so, private?

packages/vsce/src/resources/SessionHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.session = new CICSSession(profile);
+    this.sessions.set(profileName, this.session);
```
AndrewTwydell, line 35:
```suggestion
    const newSession = new CICSSession(profile);
    this.sessions.set(profileName, newSession);
```

packages/vsce/tsup.config.json
AndrewTwydell, line 1: No changes here thanks 😄

AndrewTwydell (commented):
I've been a bit picky here but I really like this idea! 🎉

packages/vsce/tsup.config.json
almightyrush: Aahh, Will revert it. 
Thanks 😛!!

AndrewTwydell (approved):
Looking gooood! 👍🏼

packages/vsce/src/trees/CICSSessionTree.ts
```diff
@@ -39,17 +39,8 @@ export class CICSSessionTree extends TreeItem {
...
-      storeCookie: true,
-      tokenType: imperative.SessConstants.TOKEN_TYPE_LTPA,
-      hostname: this.profile.profile!.host,
-      port: Number(this.profile.profile!.port),
-      user: this.profile.profile!.user || "",
-      password: this.profile.profile!.password || "",
-      rejectUnauthorized: this.profile.profile!.rejectUnauthorized,
-      protocol: this.profile.profile!.protocol,
-    });
+    SessionHandler.getInstance().removeSession(this.profile.name);
```
enamkhan, line 42: Maybe we could have reset method the encapsulates the removing and creating a new session. Although this is not a blocker.

PR #335, opened 2025-06-18
Initial command to launching resource inspector by resource name and resource type

enamkhan:
**What It Does**
This is a command to launch resource inspector by just providing the name of the resource and the type, although currently this is set to hidden.

**How to Test**
Manually tested to make sure this integration point works.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/commands/inspectResourceCommand.ts
+++ b/packages/vsce/src/commands/inspectResourceCommand.ts
@@ -9,13 +9,15 @@
  *
  */
 
-import { commands, ExtensionContext } from "vscode";
+import { commands, ExtensionContext, window } from "vscode";
 import { ResourceInspectorViewProvider } from "../trees/ResourceInspectorViewProvider";
 import { getFocusRegion } from "./setFocusRegionCommand";
 import { IFocusRegion } from "./IFocusRegion";
 import { Resource, ResourceContainer } from "../resources";
 import { IResource, IResourceMeta } from "../doc";
 import { SupportedResources } from "../model";
+import { CICSLogger } from "../utils/CICSLogger";
+import { CICSMessages } from "../constants/CICS.messages";
 
 export function getInspectResourceCommand(context: ExtensionContext) {
   return commands.registerCommand("cics-extension-for-zowe.inspectResource", async (resourceName: string, resourceType: string) => {
@@ -23,6 +25,14 @@ export function getInspectResourceCommand(context: ExtensionContext) {
 
     if (focusRegion) {
       const type = getResourceType(resourceType);
+
+      if (!type) {
+        // Error if resource type not found
+        CICSLogger.error(CICSMessages.CICSResourceTypeNotFound.message);
+        window.showErrorMessage(CICSMessages.CICSResourceTypeNotFound.message);
+        return;
+      }
+
       const resourceContainer = new ResourceContainer<IResource>(type);
       resourceContainer.setCriteria([resourceName]);
       const resources:[Resource<IResource>[], boolean] = await resourceContainer.loadResources(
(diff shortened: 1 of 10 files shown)


packages/vsce/src/commands/inspectResourceCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+
+      const resourceViewProvider = ResourceInspectorViewProvider.getInstance(context.extensionUri);
+      const enbededWebview = resourceViewProvider?._manager?._view;
+
+      // Will only have one resource
+      const resource:Resource<IResource>[] = resources[0];
+      resourceViewProvider.reloadData( {
+            resource: resource[0],
+            meta: resourceContainer.getMeta(),
+          }, enbededWebview);
```
AndrewTwydell, line 42: I'm wondering why we need to pass the enbededWebview paramter (I know that wasn't added in this PR but I've noticed now..!)... It pulls a value out of resourceViewProvider and passes it back in. Should just reference it straight from the method???

enamkhan: Will change in a follow up pr.

packages/vsce/src/commands/inspectResourceCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      if (!type) {
+        // Error if resource type not found
+        CICSLogger.error(CICSMessages.CICSResourceTypeNotFound.message);
+        window.showErrorMessage(CICSMessages.CICSResourceTypeNotFound.message);
+        return;
+      }
+
+      const resourceContainer = new ResourceContainer<IResource>(type);
+      resourceContainer.setCriteria([resourceName]);
```
AndrewTwydell, line 41: We don't check resourceName is actually a value. Not stop-ship for now but worth considering some validation in your follow-up.

packages/vsce/src/commands/inspectResourceCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+      commands.executeCommand("setContext", "zowe.vscode-extension-for-zowe.showResourceInspector", true);
+      commands.executeCommand("workbench.view.extension.inspector-panel");
+    }
+  });
+
+  function getResourceType(resourceName: string): IResourceMeta<IResource> {
+    const types: IResourceMeta<IResource>[] = SupportedResources.metaResources.filter((value) => value.resourceName == resourceName);
+
+    // Should only have one
+    if (types && types.length > 0) {
```
AndrewTwydell, line 68: Typescript thing, for future reference...
```suggestion
    if (types?.length > 0) {
```

packages/vsce/src/commands/inspectResourceCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+import { IResource, IResourceMeta } from "../doc";
+import { SupportedResources } from "../model";
+import { CICSLogger } from "../utils/CICSLogger";
+import { CICSMessages } from "../constants/CICS.messages";
+
+export function getInspectResourceCommand(context: ExtensionContext) {
+  return commands.registerCommand("cics-extension-for-zowe.inspectResource", async (resourceName: string, resourceType: string) => {
+    if (!await workspace.getConfiguration().get('zowe.cics.resourceInspector')) {
+      return;
+    };
```
AndrewTwydell, line 26:
```suggestion
    }
```

AndrewTwydell (approved):

davenice (approved):
Minor suggestion

PR #340, opened 2025-06-20
Command Palette support for RI

chinmdas:
**What It Does**

- Instills Command palette support Resource Inspector
- Declares a command before using it : `cics-extension-for-zowe.inspectResource`

### DIFF
--- a/packages/vsce/package.json
+++ b/packages/vsce/package.json
@@ -238,13 +238,18 @@
           "light": "resources/imgs/toggle-resources-dark.svg",
           "dark": "resources/imgs/toggle-resources-light.svg"
         }
+      },
+      {
+        "command": "cics-extension-for-zowe.inspectResource",
+        "title": "Invoke Resource Inspector",
+        "category": "IBM CICS for Zowe Explorer"
       }
     ],
     "menus": {
       "commandPalette": [
         {
           "command": "cics-extension-for-zowe.inspectTreeResource",
-          "when": "never"
+          "when": "config.zowe.cics.resourceInspector"
         },
         {
           "command": "cics-extension-for-zowe.setFocusRegion",
--- a/packages/vsce/package.json
+++ b/packages/vsce/package.json
@@ -241,7 +241,7 @@
       },
       {
         "command": "cics-extension-for-zowe.inspectResource",
-        "title": "Invoke Resource Inspector",
+        "title": "Inspect Resource",
         "category": "IBM CICS for Zowe Explorer"
       }
     ],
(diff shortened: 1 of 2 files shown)


davenice (commented):
Just trying it out

packages/vsce/package.json
```diff
@@ -238,13 +238,18 @@
           "light": "resources/imgs/toggle-resources-dark.svg",
           "dark": "resources/imgs/toggle-resources-light.svg"
         }
+      },
+      {
+        "command": "cics-extension-for-zowe.inspectResource",
+        "title": "Invoke Resource Inspector",
```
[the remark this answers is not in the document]
chinmdas: @davenice , `cics-extension-for-zowe.inspectTreeResource` already has that title.

```
{
        "command": "cics-extension-for-zowe.inspectTreeResource",
        "title": "Inspect Resource",
 }
```

packages/vsce/src/commands/resourceInspectorViewCommand.ts
```diff
@@ -17,7 +17,17 @@ import { findSelectedNodes } from "../utils/commandUtils";
 
 export function getResourceInspectorCommand(context: ExtensionContext, treeview: TreeView<any>) {
   return commands.registerCommand("cics-extension-for-zowe.inspectTreeResource", async (node: CICSResourceContainerNode<IResource>) => {
-    const nodes = findSelectedNodes(treeview, node.getContainedResource().meta, node);
+    let meta;
+    if (!node) {
+      for (const res of [...new Set([...treeview.selection])].filter(
+        (item) => item instanceof CICSResourceContainerNode && item.getContainedResource()?.resource
+      )) {
+        meta = res.getContainedResource().meta;
```
AndrewTwydell, line 25: So right-click `Inspect Resource` command is only enabled for single resources - it's not shown for multiselect.

Here, we're enabling command palette for multiple selections, but in doing so, can only show the _last_ selected item (we're overriding the meta variable with the last selection in the list).

While this won't error, if you select a program then a local file and run it, the localfile shows in RI. If you select a localfile then a program, the program shows in RI. If you select 2 programs, the second one shows. 

Perhaps it's more consistent to infomessage and return out if there's more than 1 resource selected, seeing as we don't have the ability to show >1 resource in RI??

davenice: I think I'd go with what you've said but removing "with label" so it says:
`Multiple CICS resources selected. Resource 'DSNCUEXT' will be inspected.` 👍

chinmdas: Thanks @davenice . Rephrased information message.

AndrewTwydell (approved):

davenice (approved):

PR #341, opened 2025-06-20
Reuse resource inspector webview

AndrewTwydell:
**What It Does**
Stores WebviewView instance in the webview provider and reuses it if it's open. Prevents recreation of webview, and therefore duplicating event listeners, and improves speed when changing resources.

Previously, a new event listener was added every time a resource was inspected, which would quickly lead to performance issues.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog


packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -31,17 +36,63 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
...
-    if (this._manager) {
-      this._manager.initializeWebview(webviewView);
-    }
-    ResourceInspectorViewProvider.refreshed = true;
+  /**
+   * Method called by VS Code when a view first becomes visible.
+   * Captures and stores the webview instance we get back to reuse for it's life.
+   * 1 event listener, 1 view, refreshable content.
+   */
+  public resolveWebviewView(webviewView: WebviewView) {
```
chinmdas, line 44: Inside `resolveWebviewView` we have missed to provide implementation for `onDidDispose() `method.
Its a good to have implementation when our webview is disposed. I can think of a simple implementation like this:
```
this.webviewView.onDidDispose(() => {
      this.webviewReady = false;
      this.webviewView = undefined;
    });
```

If you can put in better onDidDispose implementation, go for it.

chinmdas (changes requested):
Functionality looks good..!!! Tested your branch thoroughly..!!
I have put in a small suggestion, Other than that, everything else LGTM..!!!

chinmdas (approved):

PR #352, opened 2025-06-26
Store focus region object in vs code settings

almightyrush:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/commands/setFocusRegionCommand.ts
+++ b/packages/vsce/src/commands/setFocusRegionCommand.ts
@@ -35,23 +35,20 @@ export function setFocusRegionCommand() {
 export async function getFocusRegion(): Promise<IFocusRegion | undefined> {
   const quickPick = Gui.createQuickPick();
   if (await isCICSProfileValidInSettings()) {
-    const { profileName, focusSelectedRegion, cicsPlex } = await getFocusRegionFromSettings();
+    const { profileName, focusSelectedRegion, cicsPlexName } = await getFocusRegionFromSettings();
     const items = [
-      { label: `Region: ${focusSelectedRegion} | CICSplex : ${cicsPlex || "NA"} | Profile: ${profileName}`, description: "Last used region" },
+      { label: `Region: ${focusSelectedRegion} | CICSplex : ${cicsPlexName || "NA"} | Profile: ${profileName}`, description: "Last used region" },
       { label: "Other CICS Region" },
     ];
     const choice = await getChoiceFromQuickPick(quickPick, "Select Region", [...items]);
     quickPick.hide();
 
-    if (!choice) {
-      Gui.infoMessage(l10n.t("No selection made. Operation cancelled."));
-      return;
-    }
+    if (!choice) return;
 
     if (choice.label != "Other CICS Region") {
       const profile = await ProfileManagement.getProfilesCache().getLoadedProfConfig(profileName);
       const session = SessionHandler.getInstance().getSession(profile);
-      return { profile, cicsPlex, session, focusSelectedRegion };
+      return { profile, cicsPlexName, session, focusSelectedRegion };
     } else {
       return await updateFocusRegion();
     }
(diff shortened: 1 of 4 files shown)


packages/vsce/src/commands/setFocusRegionCommand.ts
```diff
@@ -10,98 +10,146 @@
...
     //if plex is selected, we should show the regions from the plex
-    cicsPlex = choice.label;
     let regionInfo = await ProfileManagement.getRegionInfo(cicsPlex, session);
+    CICSLogger.info("Fetching regions for CICSplex: " + cicsPlex);
     regionInfo = regionInfo.filter((reg) => reg.cicsstate === "ACTIVE");
-    choice = await getChoiceFromQuickPick(quickPick, "Select Region", [...regionInfo.map((region) => ({ label: region.cicsname }))]);
+    choice = await getChoiceFromQuickPick(quickPick, "Select CICS Region", [...regionInfo.map((region) => ({ label: region.cicsname }))]);
     quickPick.hide();
+    if (!choice) {
+      Gui.infoMessage(l10n.t("No selection made. Operation cancelled."));
```
AndrewTwydell, line 120: We might find these get tedious so we might remove the popups and just exit quietly, but we can see. A user know they're cancelling by hitting escape so I don't think it would hurt to not show a message.

packages/vsce/src/commands/setFocusRegionCommand.ts
```diff
@@ -10,98 +10,146 @@
...
+      return;
+    }
     focusSelectedRegion = choice.label;
+    CICSLogger.info(`region set to ${focusSelectedRegion} for profile ${profileName.label} and plex ${cicsPlex || "NA"}`);
   }
-
+  setFocusRegionIntoSettings(focusSelectedRegion, profile.name, cicsPlex);
+  CICSLogger.info(`Updatingregion in settings: ${focusSelectedRegion}, profile: ${profile.name}, plex: ${cicsPlex}`);
```
AndrewTwydell, line 127:
```suggestion
  CICSLogger.info(`Updating region in settings: ${focusSelectedRegion}, profile: ${profile.name}, plex: ${cicsPlex}`);
```

packages/vsce/src/commands/setFocusRegionCommand.ts
```diff
@@ -10,98 +10,146 @@
...
-
-async function getAllCICSProfiles(): Promise<FilterDescriptor[]> {
-  const profileInfo = await ProfileManagement.getProfilesCache().getProfileInfo();
-  const allCICSProfiles = profileInfo.getAllProfiles("cics");
-
-  const allCICSProfileNames: string[] = allCICSProfiles ? (allCICSProfiles.map((profile) => profile.profName) as unknown as [string]) : [];
-  const cicsProfiles = allCICSProfileNames.map((name) => new FilterDescriptor(name));
-  return cicsProfiles;
-}
+async function updateFocusRegion(): Promise<IFocusRegion> {
```
AndrewTwydell, line 64:
```suggestion
async function updateFocusRegion(): Promise<IFocusRegion | undefined> {
```

packages/vsce/src/commands/setFocusRegionCommand.ts
```diff
@@ -10,98 +10,146 @@
...
+  }
+
+  const profileName = cicsProfiles.find((item) => item === choice);
+  const profile = await ProfileManagement.getProfilesCache().getLoadedProfConfig(profileName.label);
+  const session = SessionHandler.getInstance().getSession(profile);
 
-  if (plex) {
+  ({ cicsPlex, focusSelectedRegion, isPlex, plex } = await getPlexAndRegion(profile, cicsPlex, focusSelectedRegion, isPlex, plex, session));
+
+  if (plex && !cicsPlex) {
```
AndrewTwydell, line 90: These names are a bit confusing, wonder if we can make them a bit more readable?

eg. `cicsPlex`, `isPlex`, `plex`. Presumably `isPlex` is a boolean, but the others I couldn't be sure without digging into the code.

packages/vsce/src/commands/setFocusRegionCommand.ts
almightyrush: I had taken references from the zowe's code. 
But yes I agree with you, removed the message.

packages/vsce/src/commands/setFocusRegionCommand.ts
almightyrush: cicsPlex = cicsPlexName and plex = plexInfo.  Would this be okay ?

PR #358, opened 2025-07-01
Breadcrumb integration for RI

chinmdas:
**What It Does**
This Integrates Breadcrumb section that will display the Profile >> Plex >> Region information

**How to Test**
- Open RI via Tree
- Open RI via Command Palette

### DIFF
--- a/packages/vsce/src/doc/resources/IResourcesHandler.ts
+++ b/packages/vsce/src/doc/resources/IResourcesHandler.ts
@@ -1,0 +1,7 @@
+import { Resource, ResourceContainer } from "../../resources";
+import { IResource } from "./IResource";
+
+export interface IResourcesHandler {
+  resources: [Resource<IResource>[], boolean];
+  resourceContainer: ResourceContainer<IResource>;
+}
--- a/packages/vsce/src/trees/ResourceInspectorView.ts
+++ b/packages/vsce/src/trees/ResourceInspectorView.ts
@@ -1,0 +1,69 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { HTMLTemplate } from "@zowe/zowe-explorer-api";
+import { randomUUID } from "crypto";
+import { Uri, Webview, WebviewView } from "vscode";
+import { IResource } from "../doc";
+import { CICSResourceContainerNode } from "./CICSResourceContainerNode";
+import Mustache = require("mustache");
+
+export class ResourceInspectorView {
+  public _view?: WebviewView;
+
+  constructor(
+    private readonly extensionUri: Uri,
+    private readonly resource: CICSResourceContainerNode<IResource>
+  ) {}
+
+  initializeWebview(webviewView: WebviewView) {
+    this._view = webviewView;
+    webviewView.webview.options = {
+      enableScripts: true,
+      localResourceRoots: [this.extensionUri],
+    };
+
+    webviewView.webview.onDidReceiveMessage(async (message) => {
+      if (message.command === "init") {
+        await this.sendDataToReactApp();
+      }
+    });
+    webviewView.webview.html = this._getHtmlForWebview(webviewView.webview);
+  }
+
+  private async sendDataToReactApp() {
+    const containedResource = this.resource.getContainedResource();
+    if (this._view) {
+      await this._view.webview.postMessage({
+        command: "init",
+        data: {
+          name: this.resource.label,
+          resourceName: containedResource.meta.resourceName,
+          highlights: containedResource.meta.getHighlights(containedResource.resource),
+          resource: containedResource.resource.attributes,
+          regionName: this.resource.regionName,
+          cicsplexName: this.resource.cicsplexName,
+          profileName: this.resource.getProfileName(),
+        },
+      });
+    }
+  }
+
+  private _getHtmlForWebview(webview: Webview): string {
+    const scriptUri = webview.asWebviewUri(Uri.joinPath(this.extensionUri, "dist", "resourceInspectorPanelView.js"));
+    const nonce = randomUUID();
+
+    return Mustache.render(HTMLTemplate.default, {
+      uris: { resource: { script: scriptUri } },
+      nonce,
+    });
+  }
+}
--- a/packages/vsce/src/trees/ResourceInspectorViewProvider.ts
+++ b/packages/vsce/src/trees/ResourceInspectorViewProvider.ts
@@ -14,6 +14,7 @@ import { randomUUID } from "crypto";
 import Mustache = require("mustache");
 import { WebviewViewProvider, Uri, WebviewView, Webview } from "vscode";
 import { IContainedResource, IResource } from "../doc";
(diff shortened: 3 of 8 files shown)


packages/vsce/src/trees/ResourceInspectorView.ts
AndrewTwydell, line 1: This file is no longer used

packages/vsce/src/doc/resources/IResourcesHandler.ts
AndrewTwydell, line 1: Need a license on this file

packages/vsce/src/webviews/resource-inspector-panel/ResourceInspector.tsx
AndrewTwydell, line 1: A lot of indentation changes here so hard to see exactly what's changed. A few class names and the Breadcrumb component I see

AndrewTwydell, line 1: With performance in mind here... You're already adding an event listener and storing what we get back in `resourceInfo`. This is being set from `event.data.data`. That will also include your new `profileHandler` parameter. I suggest you store that in the same state variable and pass it to the `Breadcrumb` component, something like...

```typescript
  const [resourceInfo, setResourceInfo] = React.useState<{
    name: string;
    resourceName: string;
    highlights: { key: string; value: string; }[];
    resource: IResource;
    // new addition to store what you're sending
    profileHandler: { key: string; value: string; }[];
  }>();
```
```typescript
// passing data to component
<Breadcrumb profileHandler={resourceInfo.profileHandler} />
```

In `Breadcrumb.tsx`
```Typescript
// Adding prop to breadcrumb component with type
const Breadcrumb = ({ profileHandler }: { profileHandler: { key: string; value: string; }[]; }) => {
...
{profileHandler && profileHandler.filter(....
```

packages/vsce/src/webviews/resource-inspector-panel/Breadcrumb.tsx
```diff
@@ -0,0 +1,51 @@
...
+      setResourceInfo(event.data.data);
+    };
+    vscode.addVscMessageListener(listener);
+
+    vscode.postVscMessage({ command: "init" });
+
+    return () => {
+      vscode.removeVscMessageListener(listener);
+    };
+  }, []);
```
AndrewTwydell, line 33: If we pass it in as a prop, we won't need this lot 😄

AndrewTwydell (commented):
Couple of suggestions around passing the data but I've tested locally and it works great!

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -73,6 +76,20 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
     }
   }
 
+  public setResourceHandlerMap(resourceHandler: IResourcesHandler): ResourceInspectorViewProvider {
+    this.resourceHandlerMap = [];
+    this.resourceHandlerMap.push({ key: "profile", value: resourceHandler.resourceContainer.getProfileName().toUpperCase() });
```
davenice, line 81: I'm not sure it's a good idea to uppercase the profile name, because I reckon they can be mixed case?

image

CICSplex and Region name are both "mainframe specials" and therefore i think it's fine to uppercase them.

davenice (changes requested):
Minor change.

I think the code looks OK. The CSS is tricky but we know we'd like to move to a React component.

In general i think the breadcrumb is important for making this live so when you open resource inspector e.g. from the command palette, you can see what you just opened.

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
davenice: It is possible you tried it with profile left as natural case and didn't like the look of it. I think it is unlikely people will have multiple profiles differentiated by case so to me it is not a high priority fix, if you had done it for aesthetics 🙂

chinmdas: Thanks for the spotting @davenice.
and yes, its purely to make it look aesthetically good. However, your point is one hundred percent valid, Let me remove the uppercasing from everything. Breadcrumbs will show them things, as they are..!!

chinmdas:
@AndrewTwydell @davenice , 
I was testing my own branch for one last time and I found a bug.
While inspecting RI from _ALLPrograms_/_ALLLocalFiles_ etc the region component will be missing.
This is because from that view `node` will not pick up region information hence. we will not be able to set it from `loadResources` utility.

Screenshot 2025-07-03 at 7 00 02 AM

davenice:
Thanks Chin. The breadcrumb will make a big difference in using the RI I think - thank you!

That hacking the region name out of the description thing is a bit nasty - Hopefully we can aim to pass that metadata through in the future. You just know that one day we'll add more stuff into that description and it will break the breadcrumb - it's brittle.

It seems to work well enough to me. I have to confess that I spent the first 10 seconds clicking on them because they grew underlines when I hovered, like a link does. Is there a reason to do that? It'll be great when in the future they have an action on them.

Other than that, looks good with a thought towards tidying up sourcing the region name in the future.

davenice:
Also I think not uppercasing the profile looks fine enough now I see it for real. Thanks for doing that.

chinmdas:
> Thanks Chin. The breadcrumb will make a big difference in using the RI I think - thank you!
> 
> That hacking the region name out of the description thing is a bit nasty - Hopefully we can aim to pass that metadata through in the future. You just know that one day we'll add more stuff into that description and it will break the breadcrumb - it's brittle.
> 
> It seems to work well enough to me. I have to confess that I spent the first 10 seconds clicking on them because they grew underlines when I hovered, like a link does. Is there a reason to do that? It'll be great when in the future they have an action on them.
> 
> Other than that, looks good with a thought towards tidying up sourcing the region name in the future.

Thanks for spotting hyperlink dave. I have now removed it, Breadcrumbs will be a read-only entity now.
On reading region from description, I have observed following:
1. When we try to open RI from a proper flow i.e going into **Plex** >> **Region** >> **Resource** , node will not set any `description` in its object reference whereas it will set `regionName`.
2. When we do a similar action from Plex level going into ALL resource section, region is the only information set in `description` whereas `regionName` is not set. It also has a `label` property which has resource label. 
ex: **DFHCSD (Unenabled)(Closed)(IYCWENW1)**
`label`: **DFHCSD (Unenabled)(Closed)**
`description`:` (IYCWENW1)`

so yea, like you said, for now we are safe, in future we need to keep a tab on utilisation of description property.
Screenshot 2025-07-04 at 7 28 05 AM

davenice:
Looks great to me. @enamkhan / @almightyrush - is one of you happy to approve? Or any further comments?

enamkhan (approved):
@chinmdas I'm going to approve this on the proviso that we have an issue to refactor Breadcrumb to make into a React component as also look at how we pass props to it.

packages/vsce/src/webviews/resource-inspector-panel/Breadcrumb.tsx
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import * as React from "react";
+
+import "../css/style.css";
+
+const Breadcrumb = ({ profileHandler }: { profileHandler: { key: string; value: string }[] }) => {
```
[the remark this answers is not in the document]
AndrewTwydell: So enam is suggesting a class component. Chin has done a functional component. They are both react components. Functional are newer and actually have more capability (able to use hooks). The arguments passed in the brackets are the new props 👍

PR #373, opened 2025-07-10
quickpick with loader and using persistent storage for last uesd region

almightyrush:
**What It Does**

Fixed await on the error messages.
QuickPick will show loading when waiting for the response.
The selected region will be stores in zowe.cics.persistent (lastUsedRegion)

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/commands/setFocusRegionCommand.ts
+++ b/packages/vsce/src/commands/setFocusRegionCommand.ts
@@ -29,7 +29,7 @@ import { CICSSession } from "@zowe/cics-for-zowe-sdk";
 
 export function setFocusRegionCommand() {
   return commands.registerCommand("cics-extension-for-zowe.setFocusRegion", async () => {
-    await getFocusRegion();
+    await updateFocusRegion();
   });
 }
 


davenice (approved):

packages/vsce/src/commands/setFocusRegionCommand.ts
```diff
@@ -28,7 +28,7 @@ import { CICSLogger } from "../utils/CICSLogger";
 
 export function setFocusRegionCommand() {
   return commands.registerCommand("cics-extension-for-zowe.setFocusRegion", async () => {
-    await updateFocusRegion();
+    await getFocusRegion();
```
AndrewTwydell, line 31: I'm not sure if this was intentional or if this was a result of manual testing. The `Set` command suggests it will update it, so `updateFocusRegion` makes sense in my mind.

packages/vsce/src/utils/PersistentStorage.ts
```diff
@@ -184,6 +194,15 @@ export class PersistentStorage {
     return this.mBundlePartSearchHistory;
   }
 
+  public getLastUsedRegion(): { regionName: string | null; cicsPlexName: string | null; profileName: string | null } {
+    return this.mlastUsedRegion as { regionName: string | null; cicsPlexName: string | null; profileName: string | null };
```
AndrewTwydell, line 198: This is going to need an interface at some point (if there isn't already?) as we'll be using this object shape all over the place!

AndrewTwydell (approved):
Couple comments to think about but looking good

PR #398, opened 2025-08-12
Switching to Playwright for e2e GUI testing

AndrewTwydell:
**What It Does**
Replaces existing e2e framework with Playwright, for ease-of-use and reliability.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/__tests__/__e2e__/specs/jvmServer.spec.ts
+++ b/packages/vsce/__tests__/__e2e__/specs/jvmServer.spec.ts
@@ -1,5 +1,5 @@
 import { expect, test } from "@playwright/test";
-import { constants, getTreeItem, prepareZoweExplorerView, resetWiremock, resetZoweExplorerView } from "../utils/helpers";
+import { constants, findAndClickText, findAndClickTreeItem, getTreeItem, prepareZoweExplorerView, resetWiremock, resetZoweExplorerView } from "../utils/helpers";
 
 test.beforeEach(async ({ page, request }) => {
   await resetWiremock(request);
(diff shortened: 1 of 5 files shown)


packages/vsce/__tests__/__e2e__/setup.sh
```diff
@@ -0,0 +1,8 @@
+
+install-extension /config/workspace/dist/cics-extension-for-zowe-3.*.vsix --extensions-dir /config/extensions
+
+mkdir -p /config/data/User
+rm /config/data/User/settings.json || true
```
AndrewTwydell, line 5: Could do a check if this file exists and run the `rm` if not, rather than having an error in the CI log

packages/vsce/__tests__/__e2e__/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
+services:
+  wiremock:
```
chinmdas, line 2: Any specific reason for this @AndrewTwydell , By doing this, we would need docker compose service to be available inside the container.
What issue did you foresee when we were bringing up wiremock container from the `workflows/main.yml`.

packages/vsce/__tests__/__e2e__/playwright.config.ts
```diff
@@ -0,0 +1,37 @@
...
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
chinmdas, line 14: Should we not have `timeout` defined here? I guess, that is a good to have feature. your thoughts?

packages/vsce/__tests__/__e2e__/playwright.config.ts
```diff
@@ -0,0 +1,37 @@
...
+  testDir: "./specs/",
```
chinmdas, line 15: Also, having **expect timeout** would be a good to have feature, i.e the maximum time expect() should wait for condition to be satisfied.

packages/vsce/package.json
```diff
@@ -779,15 +779,20 @@
...
-    "test:e2e:setup": "echo \"Using global config home: $ZOWE_CLI_HOME\" && rm -rf $ZOWE_CLI_HOME && mkdir -p $ZOWE_CLI_HOME",
-    "test:e2e:run": "tsc --project __tests__/test-tsconfig.json && extest setup-and-run '__tests__/lib/**/*.test.js' -s .cics -e .cics -c max -o __tests__/__e2e__/settings.json -i --mocha_config __tests__/__e2e__/.mocharc.js",
-    "test:e2e": "export ZOWE_CLI_HOME=`pwd`/.e2e-global-settings && npm run test:e2e:setup && npm run test:e2e:run",
+    "test:e2e:setup": "cd __tests__/__e2e__ && docker compose up -d && docker exec code-server bash /config/workspace/setup.sh",
+    "test:e2e:teardown": "cd __tests__/__e2e__ && docker compose down",
+    "test:e2e:run": "npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1",
+    "test:e2e": "npm run test:e2e:setup && npm run test:e2e:run && npm run test:e2e:teardown",
+    "test:e2e:dev": "npm run test:e2e:setup && npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1 --ui && npm run test:e2e:teardown",
+    "test:e2e:headed": "npm run test:e2e:setup && npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1 --headed && npm run test:e2e:teardown",
```
chinmdas, line 787: when will `test:e2e:headed` be called? should we not check what kinda docker image we are running inside and then configure `test:e2e` ?

packages/vsce/package.json
```diff
@@ -779,15 +779,20 @@
...
     "package": "vsce package --dependencies --allow-star-activation && node ../../scripts/mv-pack.js cics-extension-for-zowe vsix",
-    "test:e2e:setup": "echo \"Using global config home: $ZOWE_CLI_HOME\" && rm -rf $ZOWE_CLI_HOME && mkdir -p $ZOWE_CLI_HOME",
-    "test:e2e:run": "tsc --project __tests__/test-tsconfig.json && extest setup-and-run '__tests__/lib/**/*.test.js' -s .cics -e .cics -c max -o __tests__/__e2e__/settings.json -i --mocha_config __tests__/__e2e__/.mocharc.js",
-    "test:e2e": "export ZOWE_CLI_HOME=`pwd`/.e2e-global-settings && npm run test:e2e:setup && npm run test:e2e:run",
+    "test:e2e:setup": "cd __tests__/__e2e__ && docker compose up -d && docker exec code-server bash /config/workspace/setup.sh",
+    "test:e2e:teardown": "cd __tests__/__e2e__ && docker compose down",
+    "test:e2e:run": "npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1",
+    "test:e2e": "npm run test:e2e:setup && npm run test:e2e:run && npm run test:e2e:teardown",
+    "test:e2e:dev": "npm run test:e2e:setup && npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1 --ui && npm run test:e2e:teardown",
```
chinmdas, line 786: when will `test:e2e:dev` be called? should we not check what kinda docker image we are running inside and then configure `test:e2e` ?

chinmdas (changes requested):
Hi @AndrewTwydell ,

Mostly this looks great, I have some minor suggestions, have a look..!!!!!

packages/vsce/__tests__/__e2e__/docker-compose.yaml
AndrewTwydell: We need to bring up a code-server container as well, so docker is a requirement anyway - it is installed in the GH Workflow images that the job runs on by default. 

Using a docker-compose file means we can bring them up and down together. Also, when we used GH services for Wiremock, it needed to be restarted AFTER we cloned the code so we could use our mapping files. This way, we can start it when we want, so we know the mappings are already there. 

I appreciate adding Docker as a requirement for testing is a downside, but it's required to run code-server so I'm thinking we embrace it!

packages/vsce/__tests__/__e2e__/playwright.config.ts
AndrewTwydell: We could do, I was happy using the default value of 5000 ms (5 seconds) - thoughts?? Figured that was a reasonable time to wait, seeing as our requests are all fairly immediate.

packages/vsce/package.json
davenice: i think we could tweak the names. maybe make it clear that both are for local/dev use, but one is interactive? i suspect we can't have comments in the file :-(

maybe

`dev-interactive` for dev
`dev` for headed?

i wondered about non-interactive, but it's a bit clunky

packages/vsce/__tests__/__e2e__/playwright.config.ts
chinmdas: I was thinking of this _(let me know if this seems too much)_ :
```
{
   timeout: 30 * 1000, //Maximum time one test can run for
   expect: {
       timeout: 5000 //Maximum time expect() should wait for condition to be satisfied
   }
}
```

packages/vsce/package.json
davenice: i think it would be good to update the names in some way and run `npm run package` as part of starting the e2e tests?

i think this optimises for the case where the developer makes some changes, runs a build (`npm run build`) and then wants to try the e2e tests.

davenice (changes requested):

packages/vsce/__tests__/__e2e__/playwright.config.ts
AndrewTwydell: Added!

PR #411, opened 2025-09-10
Add User Agent header to VSCE requests

### LINKED ISSUE #412: Add User-Agent header to each CMCI connection from the VS Code extension
The CMCI back-end in CICS has an allowlist feature: https://www.ibm.com/docs/en/cics-ts/5.5.0?topic=sm-defining-client-allowlist-cmci-jvm-server

Customers use this to enforce rules within their org for instance around the version that are allowed to connect or specific versions that must not connect. Perhaps this is because they choose to run acceptance tests on specific versions of the CMCI client.

Passing a User-Agent header is transparent to everybody but allows the allowlist to be configured to allow versions of the VS Code extension.

AndrewTwydell:
**What It Does**
Adds the CICS extension and Zowe Explorer versions to the User Agent header of each REST request to enable CMCI AllowList compatibility.

Resolves #412

**How to Test**
View the CMCI request headers and see the following included
```
User-Agent: zowe.cics-extension-for-zowe/3.11.0 zowe.vscode-extension-for-zowe/3.3.0
```

Also run the `Copy User Agent Header` command to see the string copied to clipboard.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/vsce/__tests__/__e2e__/wiremock/mappings/CICSplex.json
AndrewTwydell, line 1: Adding the User-Agent header as a requirement for these mappings means, if the header is not present in the request, the e2e tests will not find the mapping for CICSplexes.

PR #421, opened 2025-09-17
Soft removal of "Show Attributes"

chinmdas:
**What It Does**
Removes Show Attributes context menu option from tree views and makes Inspect Resource an permanent right click menu option.

Currently, we have not performing hard delete, we'll use this property if we want to enable Show Attributes.
`"config.zowe.cics.showAttributes": true`

I am not labelling this with **NO Changelog** till we are ready to publicly announce this.

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog


Khushboo-Sharma-110597 (approved):

AndrewTwydell (approved):
Looking good

davenice:
I like the idea of a soft removal for now! Let's just check in with Ross from Z Xplore first? Also I agree that it ought to have a change-log that explains that the "Show Attributes" option is renamed to "Inspect Resource" and there is likely some readme updating needed that you could put in this PR.

davenice:
Here is the part in the readme that mentions show attributes: https://github.com/zowe/cics-for-zowe-client/blob/8ae2dc95b1c0a4aa30f59f5fba55ccc38df0a442/packages/vsce/README.md?plain=1#L215

If this PR goes in as-is, then your average user will experience Show Attributes being removed and Inspect Resource being added. So we should have it in the change-log and update the readme to have a coherent experience.

chinmdas:
> Here is the part in the readme that mentions show attributes:
> 
> https://github.com/zowe/cics-for-zowe-client/blob/8ae2dc95b1c0a4aa30f59f5fba55ccc38df0a442/packages/vsce/README.md?plain=1#L215
> 
> If this PR goes in as-is, then your average user will experience Show Attributes being removed and Inspect Resource being added. So we should have it in the change-log and update the readme to have a coherent experience.

@davenice , I have updated readme file. please have a look and suggest if we need further edits on readme.

packages/vsce/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the "cics-extension-for-zowe" extension will be documented in this file.
 
+## Recent Changes
+
+- Remove show attributes from tree menu [#422](https://github.com/zowe/cics-for-zowe-client/issues/422)
```
AndrewTwydell, line 7: I believe we should have a change-log that explains that the "Show Attributes" option is renamed to "Inspect Resource" as this most accurately describes to the user the change. Bare in mind users have not yet seen Inspect Resource in the context menu. Something like...

"Replaced the Show Attributes option with Inspect Resource in resource context menu, leveraging the updated CICS Resource Inspector."

packages/vsce/README.md
```diff
@@ -212,14 +212,6 @@ View resource information without using the CICS tree. Access the command palett
 <img src="/packages/vsce/docs/images/resource-inspector.gif" alt="CICS Resource Inspector" width="800px"/>
 </p>
 
-### Show attributes
```
AndrewTwydell, line 215: Removing this block will prevent a user getting confused and mixed messages - do we want to replace this with a block about Resource Inspector? I'll tag in @davenice and @jt-nti who have been thinking about removing a lot of the detailed feature explanations from the Readme. 

Perhaps removing this is enough for this change, in favour of a cutdown readme in the future?

jt-nti: I think there does need to be a new Resource Inspector section to replace the outgoing Show attributes section.

If we do decide to cut down the readme, the content will go somewhere and I don't think abandoning it in the mean time is the right thing to do.

PR #437, opened 2025-10-02
Move resource interfaces to VSCE-API package

AndrewTwydell:
**What It Does**
Moves the CICS resource interface files into the VSCE-API package to be available to extenders registering actions when that is exposed.

This uses the same interfaces, but imported from a different place. It also specifies in each whether they have `status`, `enablestatus`, or neither, as previously it was assumed they had `enablestatus`.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce-api/CHANGELOG.md
+++ b/packages/vsce-api/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Zowe Explorer for IBM CICS Transaction Server API wil
 
 ## Recent Changes
 
-- Add CICS resource interfaces containing specific attributes.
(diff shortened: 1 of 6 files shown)


packages/vsce-api/src/interfaces/resources/IResource.ts
AndrewTwydell, line 1: WithStatus and WithEnableStatus extend IResource as it's not consistent which status attribute is used across all resource types. The individual resources then extend the correct subinterface

davenice (approved):

Issue #433, opened 2025-10-01
Resource Inspector not refreshing after NewCopy


AndrewTwydell:
**Describe the bug**

When a program is inspected from the command palette, and a new copy is run from the action menu, the new copy count is not updated.

**Expected and actual results**

The resource is updated and the new copy count increases.

**Additional context**

This works when a resource is inspected from the tree - just not from the command palette.

https://github.com/user-attachments/assets/8c070709-5e12-414d-b0fa-e66c93758edb

https://github.com/user-attachments/assets/4a9ccd10-e551-4343-b4bc-7e92130f1579

AndrewTwydell:
After investigating, this has not worked in either of the versions that include the RI actions.

PR #450, opened 2025-10-10
Add test for refresh button in resource inspector

davenice:
**What It Does**
Adds a test for the refresh button in the resource inspector - new copies in the tree, verifies the RI hasn't updated, refreshes RI and verifies the new copy count has incremented.

This is a precursor to adding a test for this bug: https://github.com/zowe/cics-for-zowe-client/issues/433 - where we use new copy from the "..." menu and verify that the refresh occurs automatically.

No externals change - just a test.

Where hitting save has formatted the file, I've let that happen.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/__tests__/__e2e__/specs/resourceInspector.spec.ts
+++ b/packages/vsce/__tests__/__e2e__/specs/resourceInspector.spec.ts
@@ -95,14 +95,12 @@ test.describe("Resource Inspector tests", async () => {
 
     // Now check resource inspector hasn't updated
     await getResourceInspector(page).locator("#resource-title").waitFor();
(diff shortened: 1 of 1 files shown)


packages/vsce/__tests__/__e2e__/utils/helpers.ts
```diff
@@ -87,6 +88,11 @@ export const findAndClickTreeItem = async (page: Page, label: string, button: "l
   await getTreeItem(page, label).click({ button });
 };
 
+export const waitForNotification = async (page: Page, string: string) => {
```
enamkhan: string: string seems odd but minor.

PR #458, opened 2025-10-16
Externalise Strings

harshithajahagirdar:
**What It Does**
Externalise strings

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


.vscode/launch.json
AndrewTwydell: This never worked for me locally 🤔 Do you have your development VS Code in English and then run this debug task and the extension host VS Code starts in French???

packages/vsce/l10n/bundle.l10n.json
```diff
@@ -0,0 +1,40 @@
...
+  "Force Purge": "Force Purge",
+  "Kill": "Kill",
+  "Choose how to purge tasks while disabling the JVM server": "Choose how to purge tasks while disabling the JVM server",
+  "Disabling {0} of {1}": "Disabling {0} of {1}",
+  "Something went wrong when performing a disable - {0}": "Something went wrong when performing a disable - {0}",
+  "No CICS JVMEndpoint selected": "No CICS JVMEndpoint selected",
+  "${handler.dialogs.message} '${resource.jvmendpoint}' (${i + 1} of ${nodes.length}))": "${handler.dialogs.message} '${resource.jvmendpoint}' (${i + 1} of ${nodes.length}))",
+  "Something went wrong while disabling JVM Endpoint {0}\n\n{1}": "Something went wrong while disabling JVM Endpoint {0}\n\n{1}",
+  "No CICS Bundles selected": "No CICS Bundles selected",
+  "Disabling ${nodes.indexOf(node) + 1} of ${nodes.length}": "Disabling ${nodes.indexOf(node) + 1} of ${nodes.length}"
```
AndrewTwydell: Same as above

packages/vsce/src/commands/closeLocalFileCommand.ts
```diff
@@ -19,26 +20,34 @@ export function getCloseLocalFileCommand(tree: CICSTree, treeview: TreeView<any>
...
+    const picked = await window.showInformationMessage(
+      vscode.l10n.t("Choose one of the following for the file busy condition"),
+      ...BUSY_CHOICES.map((c) => c.label)
     );
-    if (!busyDecision) {
+    if (!picked) {
       return;
     }
 
+    const busyDecision = BUSY_CHOICES.find((c) => c.label === picked)?.id ?? "WAIT";
```
AndrewTwydell: Adding in a `Array.find` can add complexity and additional compute. You could use an object in cases like this, something like this may work...

```
const busyChoices = {
  [vscode.l10n.t("Wait")]: "WAIT",
  [vscode.l10n.t("No Wait")]: "NOWAIT",
  [vscode.l10n.t("Force")]: "FORCE",
}

const picked = await window.showInformationMessage(
    vscode.l10n.t("Choose one of the following for the file busy condition"),
    ...Object.keys(busyChoices)
);

if (!picked) {
    return;
}

const busyDecision = busyChoices[picked];
```

packages/vsce/src/commands/disableResourceCommand.ts
```diff
@@ -10,18 +10,18 @@
  */
 
 import { IBundle, IJVMEndpoint, IJVMServer, ILibrary, ILocalFile, IProgram, IResource, ITransaction } from "@zowe/cics-for-zowe-explorer-api";
-import { commands, TreeView, window } from "vscode";
+import * as vscode from "vscode";
+import { TreeView, commands, window } from "vscode";
```
AndrewTwydell: Same here about the import

```suggestion
import { TreeView, commands, window, l10n } from "vscode";
```

packages/vsce/package.json
```diff
@@ -12,6 +12,7 @@
   "engines": {
     "vscode": "^1.79.0"
   },
+  "l10n": "./dist",
```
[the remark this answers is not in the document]
harshithajahagirdar: After I finish the changes, I will switch this to ./l10n and try it.

.vscode/launch.json
harshithajahagirdar: I change the VS Code language from English to French, then launch the Extension Development Host by starting the debugger.

packages/vsce/cics-extension-for-zowe-3.13.2/extension/package.json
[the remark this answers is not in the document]
harshithajahagirdar: I think I pushed it by mistake.

packages/vsce/src/commands/disableResourceCommand.ts
harshithajahagirdar: Done

packages/vsce/src/commands/closeLocalFileCommand.ts
harshithajahagirdar: Changed

.vscode/launch.json
AndrewTwydell: Right okay, so this additional block in launch.json doesn't actually do anything??

packages/vsce/l10n/bundle.l10n.fr.json
davenice: let's not check translated files in when you are ready with an initial PR. they'll come from artifactory.

packages/vsce/src/commands/enableResourceCommand.ts
```diff
@@ -61,37 +60,37 @@ export const getEnableResourceCommands = (tree: CICSTree, treeview: TreeView<CIC
...
-      pollCriteria: (response) => response.records?.cicsjvmserver?.enablestatus.toUpperCase() === "ENABLED"
+      pollCriteria: (response) => response.records?.cicsjvmserver?.enablestatus.toUpperCase() === "ENABLED",
     });
   };
 
   const enableJVMEndpoint = async (treeNode: CICSResourceContainerNode<IJVMEndpoint>) => {
     const nodes = findSelectedNodes(treeview, JVMEndpointMeta, treeNode);
     if (!nodes || !nodes.length) {
-      await window.showErrorMessage("No CICS JVMEndpoint selected");
+      await window.showErrorMessage(l10n.t("No CICS JVMEndpoint selected"));
```
davenice:
```suggestion
      await window.showErrorMessage(l10n.t("No CICS JVM Endpoint selected"));
```

packages/vsce/src/commands/getFilterPlexResources.ts
```diff
@@ -32,81 +68,67 @@ export function getFilterPlexResources(tree: CICSTree, treeview: TreeView<any>)
...
     await treeview.reveal(chosenNode, { expand: true });
     const plexProfile = plex.getProfile();
-    let resourceToFilter: any;
-    if (plexProfile.profile.regionName && plexProfile.profile.cicsPlex) {
-      resourceToFilter = await window.showQuickPick(["Programs", "Local Transactions", "Local Files", "Tasks", "Libraries"]);
-    } else {
-      resourceToFilter = await window.showQuickPick(["Regions", "Programs", "Local Transactions", "Local Files", "Tasks", "Libraries"]);
+
+    const RESOURCE_CHOICES_BASE = [
+      { id: "Programs", label: l10n.t("Programs") },
```
davenice: Working back from getCmciKeyForLabel, i think we should avoid this.

Can we list the metas for the resources we want to enable, e.g. `ProgramMeta`, then we can use `<metafile>.resourceName` as the ID and `<metafile>.humanReadableNameSingular` as the display value.

.vscode/launch.json
davenice: I have never got it working. From what @harshithajahagirdar says, we can just use the normal launch config and this one can be deleted :-)

packages/vsce/src/commands/getFilterPlexResources.ts
```diff
@@ -9,13 +9,49 @@
...
 import { CICSRegionsContainer } from "../trees";
 import { CICSRegionTree } from "../trees/CICSRegionTree";
 import { CICSTree } from "../trees/CICSTree";
-import { getPatternFromFilter } from "../utils/filterUtils";
 import PersistentStorage from "../utils/PersistentStorage";
-import { CicsCmciConstants } from "@zowe/cics-for-zowe-sdk";
+import { getPatternFromFilter } from "../utils/filterUtils";
+
+/* helpers to reduce complexity and centralize mappings */
+function getCmciKeyForLabel(label: string): string | undefined {
```
davenice: then we get rid of this method - it's duplicating the kind of approach we already have in the meta files.

packages/vsce/src/commands/inquireTransaction.ts
```diff
@@ -9,36 +9,37 @@
...
       return;
     }
 
     let transactionTree: CICSResourceContainerNode<ITransaction> | undefined;
     const label = nodes[0].getParent().label;
+    const ALL_TASKS_LABEL = l10n.t("All Tasks");
 
     //if the label is All Tasks, we need to get the transaction tree from the regions node
-    if (label === "All Tasks") {
+    if (label === ALL_TASKS_LABEL) {
```
davenice: Probably we don't gain much by pulling this out to a constant - it means you have to look around to see what it was and where it came from.

packages/vsce/src/commands/inspectResourceCommandUtils.ts
```diff
@@ -233,13 +236,11 @@ async function getEntryFromInputBox(resourceType: IResourceMeta<IResource>): Pro
 
     validateInput: (value: string): string | undefined => {
       const maxLength = resourceType.maximumPrimaryKeyLength ?? constants.MAX_RESOURCE_NAME_LENGTH;
-      const tooLongErrorMessage = CICSMessages.CICSInvalidResourceNameLength.message.replace("%length%", `${maxLength}`);
+      const tooLongErrorMessage = l10n.t("Resource name must be at most {0} characters", maxLength);
```
davenice: I see that we're stopping using several bits from CICSMessages - this is fine, let's remove them from there now to tidy up.

packages/vsce/src/commands/setCICSRegionCommand.ts
```diff
@@ -19,147 +19,205 @@ import { FilterDescriptor } from "../utils/filterUtils";
 import * as regionUtils from "../utils/lastUsedRegionUtils";
 import { InfoLoaded, ProfileManagement } from "../utils/profileManagement";
 
+/* Localized templates / labels */
+const LAST_USED_TEMPLATE = (region: string, plex: string, profile: string) =>
```
davenice: Discussion point :-)

I'm not sure I like these all being moved up here.
The reason is:
1. I think it makes the code harder to read because you have a big clump of stuff here and then as you read further down you're wondering about what's in the constant
2. The VS Code localisation is very lightweight and only adds 8 characters to each string `l10n.t(`...`)` so it's not much of a problem to keep it in line.

I think with Java code we would often extract string contants like this - I'm not convinced it's what people do in TypeScript.

packages/vsce/src/utils/CICSLogger.ts
```diff
@@ -16,21 +16,25 @@ import { extensions, l10n, LogLevel, LogOutputChannel, window } from "vscode";
...
   private static logOutputChannel: LogOutputChannel;
 
-  public static initialize() {
+  public static initialize(packageJSON?: any) {
     CICSLogger.logOutputChannel = window.createOutputChannel(l10n.t("Zowe Explorer for IBM CICS TS"), { log: true });
 
     CICSLogger.info(l10n.t("Initialized logger for Zowe Explorer for IBM CICS TS"));
 
-    const packageJSON = extensions.getExtension("zowe.cics-extension-for-zowe").packageJSON;
+    // Prefer packageJSON passed from activate(context).
```
davenice: Do you have a reference for this approach? I'm not very keen.

When I had a play with this, a while back I hit a weird issue where the string "Zowe Explorer for IBM CICS TS log level: {0}" caused an exception. Is that the problem you hit?

If so, then I suggest for now we don't externalise this string with a {0} in it. Either we do a concatenation, or we just don't translate it.

That is, unless we can explain exactly why we're doing this. It looks like the kind of change that might break other things.

packages/vsce/package.json
```diff
@@ -824,7 +826,9 @@
     "test:e2e": "npm run test:e2e:setup && npm run test:e2e:run && npm run test:e2e:teardown",
     "test:e2e:dev:interactive": "npm run test:e2e:setup && npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1 --ui && npm run test:e2e:teardown",
     "test:e2e:dev": "npm run test:e2e:setup && npx playwright test -c __tests__/__e2e__/playwright.config.ts --workers=1 --headed && npm run test:e2e:teardown",
-    "pretty": "prettier --write \"src/**/*.ts\" \"__tests__/**/*.ts\""
+    "pretty": "prettier --write \"src/**/*.ts\" \"__tests__/**/*.ts\"",
+    "l10n:export": "mkdir -p dist && npx vscode-l10n-dev export \"src/**/*.ts\" --outDir l10n",
```
davenice: Should this be this?

```suggestion
    "l10n:export": "mkdir -p l10n && npx vscode-l10n-dev export \"src/**/*.ts\" --outDir l10n",
```

davenice (changes requested):
Hi Harshitha - this is coming on well, thanks.

I've added quite a few comments that I hope don't conflict with @AndrewTwydell's!

I did comment on extracting lots of things to constants. I'm not sure it's very TypeScriptish and because externalising a string is so few extra characters I think it makes the code harder to read.

packages/vsce/package.nls.fr.json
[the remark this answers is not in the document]
harshithajahagirdar: Bundle files can be ignored because they are regenerated on every compile. package.nls.json is not generated automatically, so I added the conversion manually. We might have to deliver this.

.vscode/launch.json
harshithajahagirdar: removed

packages/vsce/src/commands/setCICSRegionCommand.ts
```diff
@@ -19,147 +19,205 @@ import { FilterDescriptor } from "../utils/filterUtils";
...
-      ]);
-      regionQuickPick.hide();
-      if (!choice) return;
-
-      regionName = choice.label;
-      CICSLogger.info(`region set to ${regionName} for profile ${profileName.label} and plex ${cicsPlexName || "NA"}`);
-    } else {
-      regionQuickPick.hide();
-      Gui.showMessage(l10n.t(`No Active Regions found in ${cicsPlexName}`));
+    const regionNameSelected = await selectRegionFromPlex(cicsPlexName, profile);
```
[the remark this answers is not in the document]
harshithajahagirdar: Yes, will test this carefully.

packages/vsce/src/commands/inquireTransaction.ts
harshithajahagirdar: changed the code to "if (label === l10n.t("All Tasks")) {".

packages/vsce/src/commands/getFilterPlexResources.ts
harshithajahagirdar: removed

packages/vsce/src/utils/CICSLogger.ts
harshithajahagirdar: Okay, I have used concatenation for now. But I have used this method on many strings as of now. Will have to change that now...

PR #459, opened 2025-10-21
Combine Local and Remote files into single tree

### LINKED ISSUE #416: Add support for remote files
The CICS tree currently represents local files (the LOCFILE table), however CPSM also presents remote files (REMFILE).

From a CICS developer perspective, they just write to a file - they don't know whether it's local or remote - so I think it makes sense to show these both within a Files element in the tree.

Similar approach to TS Queue https://github.com/zowe/cics-for-zowe-client/issues/415

AndrewTwydell:
**What It Does**
Resolves #416 
Renames `Local Files` tree to `Files`, and adds Remote files to this same tree. Different context menu options are available, and different icons, but they share a parent `Files` tree. Same behaviour under a CICSplex, showing `All Files` rather than `All Local Files`

**How to Test**
Expand `Files` tree under a region with both Local and Remote files installed, and see both appear.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/__tests__/__e2e__/specs/files.spec.ts
+++ b/packages/vsce/__tests__/__e2e__/specs/files.spec.ts
@@ -1,0 +1,64 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { expect, test } from "@playwright/test";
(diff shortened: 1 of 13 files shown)


davenice (changes requested):

packages/vsce/src/commands/inspectResourceCommandUtils.ts
```diff
@@ -62,16 +62,20 @@ export async function inspectResourceByName(context: ExtensionContext, resourceN
...
-    if (!type) {
+    if (!type || type.length === 0) {
       // Error if resource type not found
       const message = CICSMessages.CICSResourceTypeNotFound.message.replace("%resource-type%", resourceType);
       CICSLogger.error(message);
       window.showErrorMessage(message);
       return;
     }
 
+    if (type[0] === LocalFileMeta || type[0] === RemoteFileMeta) {
```
enamkhan: Don't need to change anything, but wondering if the array of local and remote should be defined such that we could do something like `filesMeta.includes(type[0])`

PR #473, opened 2025-10-29
show-bundle-in-uss-dir

Khushboo-Sharma-110597:
**What It Does**
https://github.com/zowe/cics-for-zowe-client/issues/474

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

Screenshot 2025-10-30 at 2 56 03 PM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,8 +162,8 @@ jobs:
     - name: Update Dependencies
       uses: zowe-actions/octorelease/script@v1
       env:
-        GIT_COMMITTER_NAME: zowe_robot
-        GIT_COMMITTER_EMAIL: zowe.robot@gmail.com
+        GIT_COMMITTER_NAME: ${{ secrets.ZOWE_ROBOT_USER }}
+        GIT_COMMITTER_EMAIL: ${{ secrets.ZOWE_ROBOT_EMAIL }}
         NPM_RESOLUTIONS: ${{ needs.test.outputs.npm-resolutions }}
       with:
         config-dir: .github
@@ -172,31 +172,21 @@ jobs:
     - name: Update NOTICE file
       run: bash scripts/updateNotices.sh
 
-    - name: Push Notice Changes
+    - name: Commit Notice Changes
       env:
         GIT_COMMITTER_NAME: ${{ secrets.ZOWE_ROBOT_USER }}
         GIT_COMMITTER_EMAIL: ${{ secrets.ZOWE_ROBOT_EMAIL }}
-
-        GIT_CREDENTIALS: x-access-token:${{ secrets.ZOWE_ROBOT_TOKEN }}
       run: |
         set -e
 
         echo "Set config"
         git config --global user.name $GIT_COMMITTER_NAME
         git config --global user.email $GIT_COMMITTER_EMAIL
-        git config --global credential.helper store
-        git config --get remote.origin.url
-
-        touch ~/.git-credentials
-        echo "https://$GIT_CREDENTIALS@github.com" >> ~/.git-credentials
-
         git ls-remote --heads origin $GITHUB_REF
 
         git add packages/*/NOTICE
         git diff --name-only --cached
-
         git commit -s -m "Update Notice Files [ci skip]" || true
-        git push
 
     - name: Build Source
       run: npm run build


Khushboo-Sharma-110597:
> > I was considering writing an end-to-end test for this, but since the USS tree isn’t part of our extension, I couldn’t find the API calls related to it. I don’t think it’s feasible for us to write an e2e test in this case. From what I observed, they seem to be using environment variables to load data and possibly relying on real test data. Let me know your thoughts.@AndrewTwydell @davenice
> 
> Hi Khushboo, for an e2e test we'd be mocking exactly what happens on the wire, using wiremock.
> 
> So we'd look at which URLs were requested from wiremock, and then mock those using httpie or similar to get the date. We'd want to mock the minimum of their REST API that we could get away with!
> 
> You could investigate by adding localhost:8080, or wherever your wiremock is listening, as a zosmf connection and seeing which zosmf calls wiremock complains it doesn't have a mock for. You might need to start wiremock in verbose mode to see full details, I can't quite remember!

I can set up a WireMock server for z/OSMF, but adding REST API mappings is challenging because we can’t trace the z/OSMF calls using a TCP/IP debugger—the connection uses HTTPS.
Would you like me to explore an alternative approach?

packages/vsce/__tests__/__e2e__/wiremock/__files/USSTree/zosmf-info.json
```diff
@@ -0,0 +1,85 @@
+{
+  "zos_version": "05.29.00",
+  "zosmf_port": "32070",
+  "zosmf_version": "29",
+  "zosmf_hostname": "localhost",
+  "plugins": [
```
AndrewTwydell: Are all these plugins necessary?? Can we return an empty array to reduce as much of the bloat as possible?

packages/vsce/__tests__/__e2e__/wiremock/__files/CICSBundle/bundle-summary.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<response xmlns="http://www.ibm.com/xmlns/prod/CICS/smw2int" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.ibm.com/xmlns/prod/CICS/smw2int http://localhost:8080/CICSSystemManagement/schema/CICSSystemManagement.xsd" version="3.0" connect_version="0610">
+	<resultsummary api_response1="1024" api_response2="0" api_response1_alt="OK" api_response2_alt="" recordcount="1" cachetoken="E1BAE57CD2AC8A89" />
```
AndrewTwydell:
```suggestion
	<resultsummary api_response1="1024" api_response2="0" api_response1_alt="OK" api_response2_alt="" recordcount="1" cachetoken="SHOWBUNDLETOKEN" />
```

packages/vsce/__tests__/__e2e__/wiremock/mappings/scenarios/expand-bundles-from-cics-tree.json
```diff
@@ -0,0 +1,52 @@
+{
+  "mappings": [
+    {
+      "scenarioName": "Bundle expansion in cics tree",
+      "requiredScenarioState": "Started",
+      "newScenarioState": "fetch-bundles",
```
AndrewTwydell:
```suggestion
      "newScenarioState": "fetched-bundle-summary",
```

packages/vsce/__tests__/__e2e__/wiremock/mappings/scenarios/expand-bundles-from-cics-tree.json
AndrewTwydell: I think the state naming on these could be clearer, again for the wiremock log. It makes it easier to debug

packages/vsce/src/commands/showBundleDirectoryCommand.ts
```diff
@@ -0,0 +1,104 @@
...
+        label: profileName,
+        collapsibleState: vscode.TreeItemCollapsibleState.Collapsed,
+        contextValue: "uss_session",
+        profile: profile,
+        getProfile: () => profile,
+        getProfileName: () => profileName,
+    };
+}
+
+export function showBundleDirectory(treeview: TreeView<any>) {
```
AndrewTwydell: treeview is not used
```suggestion
export function showBundleDirectory() {
```

AndrewTwydell (commented):
Few comments but looking good

packages/vsce/package.json
```diff
@@ -506,6 +515,11 @@
           "command": "cics-extension-for-zowe.showRegionLogs",
           "group": "1.CICSResourceNode.inquire"
         },
+        {
+          "when": "view == cics-view && (viewItem =~ /^CICSResourceNode.CICSBundle.(ENABLED|DISABLED).*/) &&  !listMultiSelection",
```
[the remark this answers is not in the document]
AndrewTwydell: Screenshot 2025-11-04 at 2 25 26 PM

packages/vsce/__tests__/__e2e__/wiremock/mappings/scenarios/navigate-bundle-from-cics-to-uss-tree.json
```diff
@@ -0,0 +1,52 @@
...
+        "bodyFileName": "CICSBundle/CICSBundleParts/bundle-parts-records.xml",
+        "headers": {
+          "Content-Type": "text/xml"
+        }
+      }
+    },
+    {
+      "scenarioName": "Bundle Navigation from cics to uss tree",
+      "requiredScenarioState": "fetched-bundle-parts-records",
+      "newScenarioState": "fetched-bundle-parts-records",
```
AndrewTwydell:
```suggestion
      "newScenarioState": "removed-bundle-parts-cache",
```

packages/vsce/src/commands/showBundleDirectoryCommand.ts
```diff
@@ -0,0 +1,104 @@
...
+                window.showErrorMessage(`Could not find profile ${chosenProfileName}`);
+                return;
+            }
+            const ussNode = createUSSTreeNode(bundleDir, chosenProfileName, chosenProfile);
+            //If the selected profile is hidden, it should be loaded first
+            try {
+                const sessionNode = createSessionNode(chosenProfileName, chosenProfile);
+                await commands.executeCommand("zowe.uss.fullPath", sessionNode);
+            } catch (error) {
+                CICSLogger.info(`Failed to create session node: ${error}.`);
```
almightyrush:
```suggestion
                CICSLogger.error(`Failed to create session node: ${error}.`);
```

packages/vsce/src/utils/commandUtils.ts
```diff
@@ -13,6 +13,148 @@ import { TreeView } from "vscode";
...
+      const profileBase = await fetchBaseProfileWithoutError(profile);
+      
+      if (profileBase?.name === baseForCicsProfile.name) {
+        matchingBaseProfiles.push(profile);
+      }
+    }
+    
+    if (matchingBaseProfiles.length > 0) {
+      const selectedProfile = matchingBaseProfiles[0];
+      CICSLogger.debug(`Located matching z/OS profile by base profile: ${selectedProfile.name}`);
```
almightyrush:
```suggestion
      CICSLogger.info(`Located matching z/OS profile by base profile: ${selectedProfile.name}`);
```

almightyrush (changes requested):
It looks good to me. 
Lets get this in quick, couple of minor changes related to logging as this might help to get to know what the action was performed by user.
Also I see in the unit test there are auto generated comments 😜, better to get rid of it.

I know you where testing the e2e and adding those extra commits, better to squash commits 
`git reset --soft HEAD~<numberOfCommits to squash>`

davenice:
Just having a look at the code now.

I've remove the no-changelog label because we should have a change log. Apols for the confusion!

I notice if there's a bundle where the bundledir doesn't exist, we get a 404 - the location name updates in the z/OS Unix view - but the files being shown don't. I'm not sure if there's much tidying up we can do here. I think it's the same behaviour you'd get if you tried to show a directory that doesn't exist yourself, so maybe we can call that "acceptable" :-)

Khushboo-Sharma-110597:
> Just having a look at the code now.
> 
> I've remove the no-changelog label because we should have a change log. Apols for the confusion!
> 
> I notice if there's a bundle where the bundledir doesn't exist, we get a 404 - the location name updates in the z/OS Unix view - but the files being shown don't. I'm not sure if there's much tidying up we can do here. I think it's the same behaviour you'd get if you tried to show a directory that doesn't exist yourself, so maybe we can call that "acceptable" :-)

its done.

packages/vsce/src/utils/commandUtils.ts
```diff
@@ -13,6 +13,148 @@ import { TreeView } from "vscode";
...
+        break;
+      case ZoweExplorerApiType.Jes:
+        explorerApi.getJesApi(profile);
+        break;
+      default:
+        return false;
+    }
+    return true;
+  } catch {
+    CICSLogger.error(`Profile ${profile.name} does not support connection type ${connectionType}`);
```
davenice: I don't think this needs to be an error debug, because every time I run the command to show bundle directory, this runs through every available profile to check whether they support USS.

I think this should be debug - disabled by default. Otherwise the log will be overrun with "business as usual - this CICS connection isn't a USS connection"
```suggestion
    CICSLogger.debug(`Profile ${profile.name} does not support connection type ${connectionType}`);
```

packages/vsce/src/utils/commandUtils.ts
```diff
@@ -13,6 +13,148 @@ import { TreeView } from "vscode";
...
+ *
+ * @param profile - The profile to fetch the base profile for
+ * @returns The base profile or undefined if none exists
+ */
+export async function fetchBaseProfileWithoutError(profile: IProfileLoaded): Promise<IProfileLoaded | undefined> {
+  let baseProfile = undefined;
+  try {
+    baseProfile = await ProfileManagement.getProfilesCache().fetchBaseProfile(profile.name);
+  } catch (ex) {
+    CICSLogger.error(`No base profile found for ${profile.name}`);
```
davenice: I don't think we should output this as an error, because we are using this function to check whether a base profile exists or not. If it doesn't exist, that's not an error, just "no base profile".

```suggestion
    // this isn't an error we're interested in - we were checking if a base profile existed
    CICSLogger.debug(`No base profile found for ${profile.name}`);
```

davenice:
Just catching up on my email I notice my suggestions conflict with ones you made for @almightyrush 🙈 .... I think having the benefit of running it with the changes in place, my suggestions will avoid flooding the logs but mean we can still turn on more when we need it :-)

packages/vsce/src/commands/showBundleDirectoryCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+                return;
+            }
+            const ussNode = createUSSTreeNode(bundleDir, chosenProfileName, chosenProfile);
+            //If the selected profile is hidden, it should be loaded first
+            try {
+                const sessionNode = createSessionNode(chosenProfileName, chosenProfile);
+                CICSLogger.info(`executing command: zowe.uss.fullPath for ${bundleDir}`);
+                await commands.executeCommand("zowe.uss.fullPath", sessionNode);
+            } catch (error) {
+                CICSLogger.error(`Failed to create session node: ${error}.`);
```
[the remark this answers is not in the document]
Khushboo-Sharma-110597: How do you produce this error? @davenice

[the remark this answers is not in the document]
Khushboo-Sharma-110597: After adding getSession, it was throwing an error saying t.setProfileToChoice is not a function, so I had to update it for both cases. I’ve fixed the issue — please pull the latest changes and test once.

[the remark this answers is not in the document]
AndrewTwydell: I'm now getting `[error] TypeError: e.getParent is not a function`. Did you not?

https://github.com/user-attachments/assets/69c16b0d-fd03-4583-96f8-a75a76158018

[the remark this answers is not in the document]
AndrewTwydell: If we’ve got a decent end to end test, I think it should be okay. I don’t want it to swallow any other errors, but I do agree that it’s working nicely.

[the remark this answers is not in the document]
Khushboo-Sharma-110597: I’ve wrapped the “Cannot resolve tree item” error in a debug logger, so it no longer appears in the Zowe Explorer IBM CICS TS logs. However, it still shows up in the VS Code Extension Host console I wonder if it’s being logged by VS Code’s internal platform code before our catch block is reached — I think this behaviour is unavoidable from within the extension code.
what do you think?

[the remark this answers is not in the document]
AndrewTwydell: I think you're right with ZE logging it on their end as well as us catching it... Perhaps we could ask the ZE team how they would recommend doing what you're doing? I don't think it's an uncommon usecase; to want to open a specific directory on a specific profile.

AndrewTwydell (approved):

PR #485, opened 2025-11-08
Update release workflow to avoid always pushing notice changes

t1m0thyj:
The release workflow runs every time a PR is merged into the main branch, and pushes a commit that updates notice files.

I believe the intent of the "Push Notice Changes" step was to push changes only when a new release is being produced. If that is wrong, feel free to close this PR.

Removing the `git push` call should achieve this behavior, because our Octorelease tool will run `git push` if there was a version bump. Otherwise the push will be skipped which matches the behavior of the "Update Dependencies" step.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,8 +162,8 @@ jobs:
     - name: Update Dependencies
       uses: zowe-actions/octorelease/script@v1
       env:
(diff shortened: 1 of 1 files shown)


zFernand0 (approved):
LGTM! 😋 
---
Should we merge https://github.com/zowe-actions/octorelease/pull/165 first?

PR #489, opened 2025-11-11
Address identified code smells

awharn:
**What It Does**

If a node is undefined, the property `children` will also be undefined. Since the code references `Node.children.length` multiple times, not handling this will result in an error. Add handling to prevent errors if that is the case.

Please add appropriate reviewers as necessary. 😄 

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


enamkhan (approved):
LGTM!

PR #505, opened 2025-11-21
Add links to ZE Jobs tree from Resource Inspector

Khushboo-Sharma-110597:
**What It Does**

Adds hyperlinks to job spools from attributes within resource inspector. Eg. JVM Server linking to a DD on the CICS region job.

Screenshot 2025-11-25 at 2 25 39 PM

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/vsce/src/trees/ResourceInspectorViewProvider.ts
+++ b/packages/vsce/src/trees/ResourceInspectorViewProvider.ts
@@ -251,7 +251,7 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
           window.showErrorMessage(l10n.t("Could not find any record for region {0} to show logs.", regionName));
       }
     } catch (error) {
-        CICSLogger.error(`Error showing logs for hyperlink: ${error.message}`);
+        CICSLogger.error(`Error showing logs for hyperlink. Region: ${regionName}, Error: ${error.message}`);
         window.showErrorMessage(l10n.t("Failed to show logs for region {0}: {1}", regionName, error.message));
       }
   }


packages/vsce/src/webviews/resource-inspector-panel/utils/hyperlinkUtils.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import * as React from "react";
+import { IResourceProfileNameInfo } from "@zowe/cics-for-zowe-explorer-api";
+import * as vscode from "../../common/vscode";
```
AndrewTwydell: Import this directly and then you don't have to use `any`
```suggestion
import { postVscMessage } from "../../common/vscode";
```

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -212,4 +217,52 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
       nonce,
     });
   }
+
+  /**
+   * Handles the showLogsForHyperlink request from the webview
+   * Fetches job ID and calls showRegionLogs command
+   */
+  private async handleShowLogsForHyperlink() {
```
AndrewTwydell: An issue with this method is the number of paths it can take that do not tell the user anything without hunting in the extension log...

1. An error occurs during any of it
2. No regionNode was found
3. No JobID was found for the region

We're writing to the log each time but the user would click on the hyperlink and seemingly nothing would happen.

packages/vsce/__tests__/__unit__/trees/hyperlinkRegexPattern.unit.test.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ */
+
+// Regex pattern to match //DD: followed by any characters
+const DD_PATTERN = /^\/\/DD:.+/;
+
+// Function to check if a value matches the hyperlink pattern
+function isHyperlinkableValue(value: string): boolean {
+  return DD_PATTERN.test(value);
+}
```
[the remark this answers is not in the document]
Khushboo-Sharma-110597: I did this intentionally because in our jest config has moduleFileExtensions: ["ts", "js", "json"] but doesn't include tsx and the function which i am trying to import is in tsx file, i will have to update config with tsx as well, so i decided to write unit in this way, but i can add that config and we can import it

Khushboo-Sharma-110597:
> Having a think about those issues around getting the region node from the tree because you might not have the tree, perhaps we should side-step all that (and simplify the approach, arguably!) by using runGetResource to get the Region resource directly? We know whether we've got a CICSplex + region or just a single server region, so we should be able to form the query fairly easily and it would just be one request - we could have the region name as our criteria so we only request one record (this is important because the CICSRGN table is a large one).
> 
> This would also mean that if the region was restarted after the tree was loaded we'd refresh the Job ID. Not really a massive advantage, but it's there...!

@davenice  What I understand is:
1.We first fetch the region data for the selected region using the runGetResource API.
2.Then we create a RegionNode from that data and use it accordingly.
We’re inside ResourceInspectorViewProvider, and I’m trying to understand in what situations the tree wouldn’t appear. Since we select a resource and then click on it, the tree should always be present — so when would it not show up?

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
Khushboo-Sharma-110597: I have added the error messages for user to see.

common.jest_config.ts
AndrewTwydell: Why have you added `tsx` here?

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -212,4 +219,66 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
...
+      if (response.records?.cicsregion) {
+        const regionRecords = toArray(response.records.cicsregion);
+        if (regionRecords.length > 0) {
+          const regionData = regionRecords[0];
+          
+          // Create a CICSRegionTree node from the fetched region data
+          const regionNode = this.createRegionNodeFromData(regionData, regionName, profileName);
+          
+          // Call the existing showRegionLogs command with the region node
+          commands.executeCommand("cics-extension-for-zowe.showRegionLogs", regionNode);
```
AndrewTwydell: It seems odd to create a _fake_ region node (including a fake session node) purely to pass it to the showRegionLogs command where it's only used to extract the Job ID and profile (both of which we already have here)... 

Wonder if we can be clever and use what we already have from the CICSRegion request and profile to skip past the first half of showRegionLogs, rather than having to mock a region node.

davenice:
I had a play with this and it works well. I had the same thought as Andrew about not mocking up the region node if we can avoid it.

**Perhaps a small refactor of getShowRegionLogs in showLogsCommand would allow you to pass the jobid straight in**? I do quite like the neatness of calling a command, but I do quite dislike the fake region node object :-D

In terms of the actual functionality, I tested:
1. In a CICSplex - all good
2. In a standalone region (I had to fake an dd: log) - all good
3. Where I had a region group set as my regionName in my zowe.config.json file, and I find a JVM server via the tree - all good.
4. Where I had a region group set as my regionName in my zowe.config.json file, and I find a JVM server using the command palette. In this case it doesn't work correctly because the region name in the breadcrumb was not correct.

For interest, here's how the breadcrumb appeared in scenario (4):
image

And here are the logs:
```
2025-11-26 19:45:06.247 [debug] GET - Resource [CICSRegion], REGIONNAME [ALLGRP], CICSPLEX [CICSEX62]
2025-11-26 19:45:06.730 [debug] Showing region logs for region IYCWEPL1
2025-11-26 19:45:06.731 [debug] Job ID for region: STC78624
```

However this is a bit of a niche scenario. I think the real problem here is that we've passed the wrong region name in - if that was correct (and this is a bug) then your code would work fine. I was surprised not to get an error that the region didn't exist though... I guess this is because we've provided ALLGRP which is a valid scope (region group, rather than region) which has returned a bunch of regions and IYCWEPL1 is the first in the list.
**I don't think you need to fix this issue (4) as part of this work.**

I did spot one bit of behaviour I didn't like which is that when you click the hyperlink the Resource Inspector jumps back to the top - **I think it would be good to have a look at this**.

common.jest_config.ts
Khushboo-Sharma-110597: I can't import isHyperlinkableValue function from my tsx file to unit test file, that's why i was writing this function in the unit test directly just to avoid this but you wanted me to import that function then i had to modify it, if you have any other suggestion then let me know

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -212,4 +219,66 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
...
+        } else {
+          CICSLogger.debug(`Empty region records array for ${regionName}`);
+          window.showErrorMessage(`Could not find region ${regionName} to show logs.`);
+        }
+      } else {
+        CICSLogger.debug(`No region records found for ${regionName}`);
+        window.showErrorMessage(`Could not find region ${regionName} to show logs.`);
+      }
```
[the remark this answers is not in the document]
Khushboo-Sharma-110597: @AndrewTwydell I’m not sure what you mean by combining the logs and errors. Could you please explain it in more detail?

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
Khushboo-Sharma-110597: I have moved the logic to command utils, and calling from there for showlogs and resourceViewProvider

AndrewTwydell (approved):
Looks good! 👍🏼

packages/vsce/__tests__/__e2e__/specs/hyperlinkNavigation.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    const zosmfWiremockItem = getTreeItem(page, constants.ZOSMF_PROFILE_NAME, false);
+    await expect(zosmfWiremockItem).toBeVisible();
+    
+    await zosmfWiremockItem.click();
+    await page.waitForTimeout(500);
+    await page.screenshot({ fullPage: true, path: "./__tests__/screenshots/resourceInspector/hyperlinks/5.png" });
+    
+    const jobItem = getTreeItem(page, "job1(jobid1) - ACTIVE", false);
```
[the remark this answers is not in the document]
davenice: This is stale. It is used.

packages/vsce/src/trees/ResourceInspectorViewProvider.ts
```diff
@@ -212,4 +218,41 @@ export class ResourceInspectorViewProvider implements WebviewViewProvider {
...
+          } else {
+              CICSLogger.debug(`Empty region records array for ${regionName}`);
+              window.showErrorMessage(l10n.t("Could not find region data and job id for region {0} to show logs.", regionName));
+          }
+      } else {
+          CICSLogger.debug(`No region records found for ${regionName}`);
+          window.showErrorMessage(l10n.t("Could not find any record for region {0} to show logs.", regionName));
+      }
+    } catch (error) {
+        CICSLogger.error(`Error showing logs for hyperlink: ${error.message}`);
```
davenice: i think we should put the region name in
```suggestion
        CICSLogger.error(`Error showing logs for hyperlink. Region: ${regionName}, Error: ${error.message}`);
```

will error.message lose the stack trace? in which case is there a better way to log the whole error?

davenice (approved):
Tiny changes

Issue #513, opened 2025-11-24
Externalise strings to allow them to be translated


davenice:
Externalise strings using the l10n method similar to Zowe Explorer - this will allow them to be translated.

PR #518, opened 2025-11-25
Externalise Strings: PR7. Externalise strings from other folders PR7

harshithajahagirdar:
**What It Does**

Working on https://github.com/zowe/cics-for-zowe-client/issues/513

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/vsce/__tests__/__unit__/doc/meta.jvmserver.test.ts
harshithajahagirdar: I tried the steps, but it didn't give me an updated file to push. I know it looks overwhelming, but I have just changed a single word. 
Line 11.
from: expect(label).toEqual(`JVM1 (Disabled)`);
To: expect(label).toEqual(`JVM1 (DISABLED)`);
One way I found was to use text compare.

packages/vsce/__tests__/__e2e__/specs/jvmServer.spec.ts
```diff
@@ -76,11 +76,11 @@ test.describe("JVM server tests", () => {
 
     await expect(page.getByRole("button", { name: "Phase Out", exact: true })).toBeVisible();
     await page.getByRole("button", { name: "Phase Out", exact: true }).click();
-    await expect(getTreeItem(page, `${constants.JVM_SERVER_1_NAME} (Disabled) `)).toBeVisible();
-    await expect(getTreeItem(page, `${constants.JVM_SERVER_1_NAME} (Disabled) `)).toHaveText(`${constants.JVM_SERVER_1_NAME} (Disabled)`);
+    await expect(getTreeItem(page, `${constants.JVM_SERVER_1_NAME} (DISABLED) `)).toBeVisible();
```
davenice: I don't think we should be changing these, I think we can fix the code to produce Disabled.

This is what I think we'd be expecting across the board.

I'll comment on those lines of code!

davenice (changes requested):
I think you've made things better here! Let's tweak it to stick with "Disabled", "Disabling" "Closed" etc and then I hope we're good to go.

packages/vsce/__tests__/__e2e__/specs/jvmServer.spec.ts
harshithajahagirdar: I have reverted.

packages/vsce/src/doc/meta/JVMServer.meta.ts
```diff
@@ -9,37 +9,37 @@
...
   },
 
   getLabel(resource: Resource<IJVMServer>): string {
     let label = `${resource.attributes.name}`;
-    if (resource.attributes.enablestatus.trim().toLowerCase() === "disabled") {
-      label += " (Disabled)";
+
+    if (resource.attributes.enablestatus.trim().toLowerCase() !== "enabled") {
+      const status = resource.attributes.enablestatus.trim();
+      label += ` (${status.charAt(0).toUpperCase()}${status.slice(1)})`;
```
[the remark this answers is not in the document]
harshithajahagirdar: Yup this worked. I tested a bitt too. And it works for me.

PR #530, opened 2025-12-02
Reset search string when Resource Inspector gets new resource

### LINKED ISSUE #481: Search state is persisting even after changing the Resource while performing inspect resource.
**Describe the bug**

When inspecting a resource, and then searching the attributes , and then moving to different resource , the previous searched attribute is still persisting, and the full list of attributes are not displayed.

Steps to reproduce
Right click on any resource and click on Inspect Resource.
Search for any attribute.
Go back to any other resource or any other object in the same resource, and click on Inspect Resource.

Details about the behavior:
1. Command issued: when any attribute is searched  while inspecting the resource, and then inspecting any other resource.
2. Expected results:When inspecting a resource, and then searching the attributes , and then moving to different resource , the previous searched section must be cleared, and full list of attributes should be displayed. 
3. Actual results:When inspecting a resource, and then searching the attributes , and then moving to different resource , the previous searched attribute is still persisting, and the full list of attributes are not displayed.

Image

Image

BRSinha08:
**What It Does**
When inspecting a resource, and then searching the attributes , and then moving to different resource , the previous searched attribute will not be persisting anymore and the search field will be cleared, and the full list of attributes will be displayed.

**How to Test**
Right click on any resource and click on Inspect Resource.
Search for any attribute.
Go back to any other resource or any other object in the same resource, and click on Inspect Resource.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Resolves #481


almightyrush (approved):
LGTM!

PR #540, opened 2025-12-09
Allow tree to remain expanded when refreshed

### LINKED ISSUE #472: Zowe CICS Explorer collapses tree view on refresh
**Describe the bug**

I am trying to recycle a JVM server to bring in changes in my CICS region, and in doing so I need to wait for the bundles to re-enable as they have a somewhat delayed/asynchronous lifecycle. Its easy to disable/enable the JVM server in Zowe, but how do you refresh a view to check if resources are enabled?
In the following view I can see the status of the bundle resources, but I need to know when they get to the right status. If I click the refresh icon (top right), then the view collapses back to the root of the tree, so you can't see any bundles anymore.

**Screenshots**
In the following screen shot you can see some of the bundles like HELLO are disabled and some are not. 
Image

**Expected behavior**

I would like for the refresh icon to not collapse the tree view and instead refresh the current view without modifying it.

**Desktop (please complete the following information):**

- OS: Win11
- Zowe Explorer Version: 3.3.1
- Zowe Explorer for IBM CICS Transaction Server - V10.015

AndrewTwydell:
**Based on unit-test-refactor PR branch**

**What It Does**
Resolves #472 .
When refreshing the tree, it now refreshes all expanded nodes, in the same way ZE refreshses, rather than collapsing everything and reloading the Sessions. 

**How to Test**
Expand trees, click refresh, notice resources and attributes updating.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/vsce/src/trees/CICSSessionTree.ts
+++ b/packages/vsce/src/trees/CICSSessionTree.ts
@@ -79,7 +79,7 @@ export class CICSSessionTree extends TreeItem {
 
     const configInstance = await ProfileManagement.getConfigInstance();
     if (!configInstance.getTeamConfig().exists) {
-      return;
+      return [];
     }
 
     let plexInfo: InfoLoaded[] = [];


chinmdas (approved):
Mostly Looks good to me..!!! some minor good to have suggestions, you wont need another review from my side.

Issue #500, opened 2025-11-18
Resource Inspector highlight information for all resources


BRSinha08:
**Extending highlights to all resources**
Currently, we show a minimal set of highlighted attributes (attributes we consider important) for the CICS resources in Resource Inspector. There are only a few filled out fully (local file for example). We should add any important attributes to the highlight section of the resource meta files.

**Fixing missing information from local file permissions highlight**
There are also some missing from local file's permissions highlight. We mention whether the file is readable and browsable but other permissions exist that aren't shown. e.g. addable, deletable, updatable. We should add those in. Possibly in ascending order of "severity" of action...
Browsable, Readable, Updatable, Addable, Deletable ... or similar.
Is there a better way of showing this because it will get cluttered fast? But probably this is a good first step.

PR #541, opened 2025-12-10
added information for local files permission

BRSinha08:
**What It Does**
Added  missing information from local file permissions highlight eg: UPDATABLE, ADDABLE, DELETABLE, RECOVERABLE.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Resolves issue #500

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 ## Recent Changes
 
 Enhancement: Organised resource tree children alphabetically. [#439] (https://github.com/zowe/cics-for-zowe-client/issues/439)
-Enhancement: Enhancement: Added remaining permissions attributes to CICS local file highlights. [#500](https://github.com/zowe/cics-for-zowe-client/issues/500)
+Enhancement: Added remaining permissions attributes to CICS local file highlights. [#500](https://github.com/zowe/cics-for-zowe-client/issues/500)
 
 ## `3.16.0`
 


davenice (changes requested):
Thanks Bikash.

A couple of quick things to fix.
1. I don't think we should include RECOVSTATUS - I had to look at the documentation online. I think it's not about permissions, but more about whether the file is in a state that is recoverable or not. Let's remove that attribute from the list.

2. The presentation needs to be back in line with how it was before.
Image

This is how it looks now. Notice that there are no spaces in the string other than one part way through.

The neat way to present a list like this is with spaces after the commas. So I expect to see:
`Permission: BROWSABLE, READABLE, UPDATABLE, ADDABLE, DELETABLE`

This is in line with the presentation before, where we had "READABLE, BROWSABLE".

The order of the permissions looks good to me - thank you!

packages/vsce/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 ## Recent Changes
 
 Enhancement: Organised resource tree children alphabetically. [#439] (https://github.com/zowe/cics-for-zowe-client/issues/439)
+Enhancement: Added  missing information from local file permissions highlight eg: UPDATABLE, ADDABLE, DELETABLE, RECOVERABLE. [#500](https://github.com/zowe/cics-for-zowe-client/issues/500)
```
davenice:
```suggestion
Enhancement: Added missing information from local file permissions highlight eg: UPDATABLE, ADDABLE, DELETABLE. [#500](https://github.com/zowe/cics-for-zowe-client/issues/500)
```

davenice (changes requested):
Looks good. Update changelogs (I've offered the commits you can accept if you like) and we're there.

I see RECOVSTATUS in the test data but I agree we can leave it there 👌

enamkhan (commented):
You may want to run npm pretty.

AndrewTwydell (commented):
Works nicely - couple of changelog suggestions but approving early

The wrapping of the text works fine when the longer values are present.
Image

Issue #537, opened 2025-12-05
Link to open resource documentation from error notification.


almightyrush:
1. The error notification should have a button OR link. 
    It should open the documentation showing the error codes and its description for the specific resource.
    For Example: 
    Error notification on killing JVM server 
    Should open this [URL](https://www.ibm.com/docs/en/cics-ts/6.x?topic=commands-set-jvmserver) 

2. The URLs may keep on changing, so implement unit tests to validate the URL and content in it.

PR #553, opened 2025-12-24
Add "open documentation" button to error notifications

almightyrush:
**What It Does**

Button added in the error notification which should open the documentation page for the respected resource. 
The current change is only available for the **SET RESOURCE (PROGRAMS, LIBRARY, ETC. )**  operations.

**How to Test**

Try the exception scenarios where error occurs on operations, like directly _killing the jvm_server or disabling DFHRPL_ library. 
The error notification should pop up with a button, on clicking it should redirect to the error codes page of the resource.

https://github.com/user-attachments/assets/844eb27f-58cb-4d3f-9d25-93de030559ff

Redirects to this page : [SET_JVMSERVER](https://www.ibm.com/docs/en/cics-ts/6.x?topic=commands-set-jvmserver)

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

https://github.com/zowe/cics-for-zowe-client/issues/537

### DIFF
--- a/packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
+++ b/packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
@@ -30,12 +30,7 @@ describe("Test suite to validate IBM Documentation URL", () => {
   for (const resource of resources) {
     it(`should successfully validate the documentation link for ${resource}`, async () => {
       const baseUrl = generateDocumentationURL(resource).toString(true);
-      const response = await fetch(baseUrl, {
-        method: "GET",
-        headers: {
-          Accept: "text/html",
-        },
-      });
+      const response = await fetchUrlResponse(baseUrl);
 
       expect(response.status).toBe(200);
       expect(response.url).toBe(baseUrl);
--- a/packages/vsce/__tests__/__unit__/utils/urlUtils.unit.test.ts
+++ b/packages/vsce/__tests__/__unit__/utils/urlUtils.unit.test.ts
@@ -11,53 +11,60 @@
 
 import * as vscode from "vscode";
 import { openDocumentation } from "../../../../vsce/src/utils/urlUtils";
-
-jest.mock("vscode", () => {
-  const createMockUri = (url: string): any => ({
-    _url: url,
-    with: jest.fn((change: any): any => {
-      if (change.query) {
-        return createMockUri(`${url}?${change.query}`);
-      }
-      return createMockUri(url);
-    }),
-  });
-
-  return {
-    env: {
-      openExternal: jest.fn().mockResolvedValue(true),
-    },
-    l10n: {
-      t: jest.fn((message: string) => message),
-    },
-    Uri: {
-      parse: jest.fn((url: string) => createMockUri(url)),
-    },
-  };
-});
+const uriParseSpy = jest.spyOn(vscode.Uri, "parse");
+const openExternalSpy = jest.spyOn(vscode.env, "openExternal");
 
 describe("Test suite for UrlUtils - openDocumentation", () => {
   const BASE_URL = "https://www.ibm.com/docs/en/cics-ts/6.x";
-
   beforeEach(() => {
     jest.clearAllMocks();
   });
-
   it("should open documentation for 'program' resource with correct URL", async () => {
     await openDocumentation("program");
+    expect(openExternalSpy).toHaveBeenCalledTimes(1);
+    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);
+    expect(openExternalSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ authority: "www.ibm.com", path: "/docs/en/cics-ts/6.x", scheme: "https", query: "topic=sc-set-program" })
+    );
+  });
 
-    expect(vscode.env.openExternal).toHaveBeenCalledTimes(1);
-    expect(vscode.Uri.parse).toHaveBeenCalledWith(BASE_URL);
+  it("should open documentation for 'bundle' resource with correct URL", async () => {
+    await openDocumentation("bundle");
+    expect(openExternalSpy).toHaveBeenCalledTimes(1);
+    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);
+    expect(openExternalSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ authority: "www.ibm.com", path: "/docs/en/cics-ts/6.x", scheme: "https", query: "topic=sc-set-bundle" })
+    );
+  });
 
-    const calledUri = (vscode.env.openExternal as jest.Mock).mock.calls[0][0];
-    expect(calledUri._url).toBe(`${BASE_URL}?topic=sc-set-program`);
+  it("should open documentation for 'tsqueue' resource with correct URL", async () => {
+    await openDocumentation("tsqueue");
+    expect(openExternalSpy).toHaveBeenCalledTimes(1);
+    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);
+    expect(openExternalSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        authority: "www.ibm.com",
+        path: "/docs/en/cics-ts/6.x",
+        scheme: "https",
+        query: "topic=commands-set-tsqueue-tsqname",
+      })
+    );
+  });
+
+  it("should open documentation for 'transaction' resource with correct URL", async () => {
+    await openDocumentation("transaction");
+    expect(openExternalSpy).toHaveBeenCalledTimes(1);
+    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);
+    expect(openExternalSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ authority: "www.ibm.com", path: "/docs/en/cics-ts/6.x", scheme: "https", query: "topic=commands-set-transaction" })
+    );
   });
 
   it("should open default documentation if resourcetype is undefined", async () => {
     await openDocumentation();
-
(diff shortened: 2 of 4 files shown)


packages/vsce/src/errors/urlConstants.ts
```diff
@@ -0,0 +1,27 @@
...
+  DOCPAGE: "docs",
+  ENLANGUAGE: "en",
+  CICSTS_PAGE: "cics-ts",
+  VERSION: "6.x",
+  TOPIC: "topic",
+
+  COMMANDS_SET: "commands-set-",
+  SC_SET_PROGRAM: "sc-set-program",
+  SC_SET_BUNDLE: "sc-set-bundle",
+  TSQUEUE_TSQNAME: "tsqueue-tsqname",
```
davenice: I think by extracting these to a separate place you make it harder to understand what's going on. If they were peppered in several places across the code I'd agree, but we just have that one small function that has them all in so I think you get better code clarity with the constants in here.

If there's a specific reason why it's better to do it, let's discuss!

```suggestion
```

packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { generateDocumentationURL } from "../../src/utils/urlUtils";
+
+const resources = [
```
davenice: We will absolutely need to automatically generate these.

## Why?

We see in urlUtils.ts that the doc isn't consistent and special cases are needed. There's every chance that when we add a new resource type, a special case might be needed and we forget to verify or add here.

In that case we won't know until we get the defect in, that things don't work.

## How?

I think an option here would be to add a field to the meta files called eibfnName which was like "SET JVMSERVER" or "JVMSERVER" if you want to keep your _remove the word SET_ code. Then you could loop through each of the meta files to pull out the strings to test.

Another option would be to push both the eibfnName and the topic name into the meta file 🤔 which would remove your special cases.
```
eibfnName="SET JVMSERVER"
helpTopicName="commands-set-jvmserver"
```

Happy to discuss!

davenice (changes requested):
A few suggestions - feel free to come back and discuss :-D

packages/vsce/src/utils/urlUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    case "bundle":
+      return baseUri.with({ query: `topic=${URLConstants.SC_SET_BUNDLE}` });
+
+    case "tsqueue":
+      resourceType = URLConstants.TSQUEUE_TSQNAME;
+      break;
+  }
+
+  return baseUri.with({ query: `topic=${URLConstants.COMMANDS_SET}${resourceType.toLowerCase()}` });
```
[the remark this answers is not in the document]
almightyrush: Seems cool !! updated it

chinmdas (changes requested):
One minor suggestion.... Mostly LGTM..!!!

packages/vsce/src/errors/CICSErrorHandler.ts
```diff
@@ -23,9 +25,20 @@ export function resourceNotFoundError(error?: ICICSExtensionError) {
...
-  static handleCMCIRestError(error: CICSExtensionError): Thenable<string | MessageItem> {
-    const msg = error.cicsExtensionError.errorMessage;
-    return this.notifyErrorMessage({ errorMessage: msg });
+  static handleCMCIRestError(error: CICSExtensionError, action?: MessageItem[]): Thenable<string | MessageItem> {
+    const { errorMessage: msg, resourceType } = error.cicsExtensionError;
+    const actions = resourceType && !action ? [URLConstants.OPEN_DOCUMENTATION] : action;
+    const result = this.notifyErrorMessage({ errorMessage: msg, action: actions });
+
+    if (resourceType && !action) {
+      result.then(async (selection) => {
```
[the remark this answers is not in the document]
almightyrush: So that won't be the issue, as the `result.then` will be triggered only if the button is pressed. Also the openDocumentation is async it needs `await` and so `async`.

packages/vsce/src/errors/urlConstants.ts
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { l10n } from "vscode";
+
+export const URLConstants = {
+  OPEN_DOCUMENTATION: { title: l10n.t("Open Documentation") },
```
AndrewTwydell: Is there a reason this is different, being an object with a `title` attribute?

packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
```diff
@@ -0,0 +1,72 @@
...
+describe("Test suite to validate IBM Documentation URL", () => {
+  for (const resource of resources) {
+    it(`should successfully validate the documentation link for ${resource}`, async () => {
+      const baseUrl = generateDocumentationURL(resource).toString(true);
+      const response = await fetch(baseUrl, {
+        method: "GET",
+        headers: {
+          Accept: "text/html",
+        },
+      });
```
AndrewTwydell: Is there a reason this does not use your utility method `fetchUrlResponse`?

packages/vsce/__tests__/__unit__/utils/urlUtils.unit.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  it("should open default documentation if resourcetype is undefined", async () => {
+    await openDocumentation();
+
+    const calledUri = (vscode.env.openExternal as jest.Mock).mock.calls[0][0];
+    expect(vscode.Uri.parse).toHaveBeenCalledWith(BASE_URL);
+    expect(calledUri._url).toBe(BASE_URL);
+  });
+});
```
AndrewTwydell: Because we can now manage our own version of the mocked VSCode package, you shouldn't need to do much additional mocking... I suggest something like, where you can use spies on existing mocks

The only caveat is you'll need to update our mock for `openExternal`

```
... # __mocks__/vscode.ts
    env: {
    clipboard: {
      writeText: (v: string) => { },
    },
    openExternal: jest.fn().mockResolvedValue(true),
  },
...
```

```suggestion
import * as vscode from "vscode";
import { openDocumentation } from "../../../../vsce/src/utils/urlUtils";

const uriParseSpy = jest.spyOn(vscode.Uri, "parse");
const openExternalSpy = jest.spyOn(vscode.env, "openExternal");

describe("Test suite for UrlUtils - openDocumentation", () => {
  const BASE_URL = "https://www.ibm.com/docs/en/cics-ts/6.x";

  beforeEach(() => {
    jest.clearAllMocks();
  });

  it("should open documentation for 'program' resource with correct URL", async () => {
    await openDocumentation("program");

    expect(openExternalSpy).toHaveBeenCalledTimes(1);
    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);

    expect(openExternalSpy).toHaveBeenCalledWith(expect.objectContaining({ "authority": "www.ibm.com", "path": "/docs/en/cics-ts/6.x", "scheme": "https", "query": "topic=sc-set-program" }));
  });

  it("should open default documentation if resourcetype is undefined", async () => {
    await openDocumentation();

    expect(openExternalSpy).toHaveBeenCalledWith(expect.objectContaining({ "authority": "www.ibm.com", "path": "/docs/en/cics-ts/6.x", "scheme": "https" }));
    expect(uriParseSpy).toHaveBeenCalledWith(BASE_URL);
  });
});
```

packages/vsce/src/errors/CICSErrorHandler.ts
```diff
@@ -23,9 +25,20 @@ export function resourceNotFoundError(error?: ICICSExtensionError) {
 }
 
 export class CICSErrorHandler {
-  static handleCMCIRestError(error: CICSExtensionError): Thenable<string | MessageItem> {
-    const msg = error.cicsExtensionError.errorMessage;
-    return this.notifyErrorMessage({ errorMessage: msg });
+  static handleCMCIRestError(error: CICSExtensionError, action?: MessageItem[]): Thenable<string | MessageItem> {
```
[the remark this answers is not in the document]
almightyrush: Yes, its for the future use. Consider for scenario like when the authentication fails. 
Remove the auto popping of re-entering the creds and make it available on click from the button. 
Wanted to keep this PR short so kept that for next PR 😉

packages/vsce/src/utils/urlUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+  const cicsDocHost = `${URLConstants.HOSTNAME}/${URLConstants.DOCPAGE}/${URLConstants.ENLANGUAGE}/${URLConstants.CICSTS_PAGE}/${URLConstants.VERSION}`;
+  const baseUri = Uri.parse(cicsDocHost);
+
+  if (!resourceType) {
+    return baseUri;
+  }
+
+  switch (resourceType) {
+    case "program":
+      return baseUri.with({ query: `topic=sc-set-program` });
```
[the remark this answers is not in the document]
almightyrush: Yes, I am working on it. Will put up in the next pr!

packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
almightyrush: Yes, this would be better for maintaining it. I am putting this up in the next PR. Would that be okay ?

chinmdas:
Looks good to me guys...!!!

PR #556, opened 2026-01-05
Enable CICS regions to be inspected using Resource Inspector

harshithajahagirdar:
**What It Does**

Migrate from existing Show Attributes view to the Resource Inspector component, for a consistent experience when viewing information about CICS resources and CICS regions.

**Whats in this PR?**
1. Two new commands are added inspectRegion and inspectTreeRegion, to inspect region.
2. Inspect region by name and by node.

**What is coming in the subsequent PR**
1. Removal of show attributes.
3. Adding Show Region Logs and Show SIT Parameters actions into resource inspector.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Working towards https://github.com/zowe/cics-for-zowe-client/issues/568


packages/vsce/src/webviews/resource-inspector-panel/Breadcrumb.tsx
```diff
@@ -35,30 +35,30 @@ const renderIcon = (resourceIconPath: IconPath, isDarkTheme: boolean, alt: strin
...
+    items.push(resourceContext.regionName);
+    return items;
+  }
+
   const items = [resourceContext.regionName, `${resourceName} (${humanReadableNameSingular})`];
   if (resourceContext.cicsplexName) {
     items.unshift(resourceContext.cicsplexName);
   }
   return items;
 };
```
davenice: This section of code here is a bit gnarly to understand.

You have an if statement with a return in it. Then another block after (which was pre-existing) with another return in it.

I think it would benefit from being simpler.

```suggestion
  const items: string[] = [];
  if (resourceContext.cicsplexName) {
    items.push(resourceContext.cicsplexName);
  }
  items.push(resourceContext.regionName);
  if (!isRegionType) {
    items.push(`${resourceName} (${humanReadableNameSingular})`);
  }
  return items;
};
```

PR #562, opened 2026-01-15
Remove duplicate e2e tests

AndrewTwydell:
**What It Does**
Currently there are tests that, as a first step, open the tree and assert correct resources are shown. They then continue to perform behaviour we are testing like enable/disable etc. 

However, there are also tests that just perform the existence check, which is then repeated elsewhere. We can remove the simpler tests, as the assertions as performed as part of all other tests for that resource, which will save CI time.

Any extra checks performed in deleted tests is moved to other places so no functionality is lost.

**How to Test**
Run e2e tests

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


almightyrush (approved):
Looks Good!!

chinmdas (approved):
Looks good to me..!!!!

PR #564, opened 2026-01-19
Add URLs for Get resource from the error notification.

almightyrush:
**What It Does**
The errors on the get calls like fetching the list of resources or searching with Invalid Param will have the link to the documentation in the notification. 
Updated the meta files containing the query params for SET Perform operations instead of hard coding.

**How to Test**

Try searching in programs like :  `R&^S` or something similar should throw error having button linked to specific doc.
Or any scenario where GET call throws error.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

More work on https://github.com/zowe/cics-for-zowe-client/issues/537

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -9,7 +9,6 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 - Enhancement: Open resource documentation from error notification. [#537](https://github.com/zowe/cics-for-zowe-client/issues/537)
 - Update dependencies. [#567](https://github.com/zowe/cics-for-zowe-client/pull/567)
 - Enhancement: Discard CMCI result sets when the resource node is collapsed. [#310](https://github.com/zowe/cics-for-zowe-client/issues/310)
-- Enhancement: Remove auto prompt on 401 and add "Update Credentials" button in notification. [#566] https://github.com/zowe/cics-for-zowe-client/issues/566
 
 ## `3.17.0`
 
--- a/packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
+++ b/packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
@@ -30,8 +30,8 @@ describe("Test suite to validate IBM Documentation URL", () => {
     });
   }
 
-  it(`should successfully validate IBM Documentation hompage`, async () => {
-    const baseUrl = generateDocumentationURL(undefined).toString(true);
+  it(`should successfully validate IBM Documentation URL for Get Resource`, async () => {
+    const baseUrl = generateDocumentationURL("get").toString(true);
 
     const response = await fetchUrlResponse(baseUrl);
 
@@ -40,7 +40,7 @@ describe("Test suite to validate IBM Documentation URL", () => {
     const content = await response.text();
     expect(content.length).toBeGreaterThan(0);
 
-    expect(content).toContain("IBM Documentation");
+    expect(content).toContain("GET command");
   });
 });
 
--- a/packages/vsce/src/errors/CICSErrorHandler.ts
+++ b/packages/vsce/src/errors/CICSErrorHandler.ts
@@ -10,35 +10,24 @@
  */
 
 import { Gui } from "@zowe/zowe-explorer-api";
-import { MessageItem } from "vscode";
-import errorConstants from "../constants/CICS.errorMessages";
+import { l10n, MessageItem } from "vscode";
 import { CICSLogger } from "../utils/CICSLogger";
-import { openDocumentation } from "../utils/urlUtils";
+import { generateDocumentationURL } from "../utils/urlUtils";
 import { CICSExtensionError } from "./CICSExtensionError";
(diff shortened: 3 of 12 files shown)


packages/vsce/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 
 - Enhancement: Organised profile tree alphabetically. [552](https://github.com/zowe/cics-for-zowe-client/issues/552)
 - Enhancement: Open resource documentation from error notification. [#537](https://github.com/zowe/cics-for-zowe-client/issues/537)
+- Enhancement: Remove auto prompt on 401 and add "Update Credentials" button in notification. [#566] https://github.com/zowe/cics-for-zowe-client/issues/566
```
davenice: This is the wrong one! It's snuck in :-)

packages/vsce/src/utils/errorUtils.ts
```diff
@@ -9,6 +9,42 @@
...
 export function getErrorCode(error: any) {
   return error.mDetails?.errorCode || error.response?.status;
 }
+
+export function getHelpTopicNameFromMetas(resourceType?: string): { queryParam: string; fragment: string } | undefined {
+  if (resourceType === URLConstants.GET_RESOURCE) {
+    return { queryParam: URLConstants.GET_COMMAND_URI, fragment: URLConstants.GET_COMMAND_URI_FRAGMENT };
+  }
+  const result = getResourceTypeAndHelpTopic(resourceType);
+  if (!result || !result.queryParam || !result.fragment) {
```
davenice: It looks around the place like the fragment (anchor?) is optional.

But here it looks like if the anchor isn't provided we won't return anything.

Do you need to separate the topic and anchor or would everything be simpler if you just said in the metas:
`helpTopicForSet="command-set-urimap#dfha8_seturimap__title__6"`

Do they ever need to be separate? 🤔

almightyrush: Nice catch!! 
I will remove that condition. 
Can't keep the queryparam and topic in one variable. as it gets encoded when using URI.build().
What I found is that the URI has `fragment` attributes which adds **`#`** and appends the text

packages/vsce/src/utils/urlUtils.ts
```diff
@@ -26,17 +27,13 @@ export function generateDocumentationURL(resourceType?: string): Uri {
...
 
-    case "bundle":
-      return baseUri.with({ query: `topic=sc-set-bundle` });
-
-    case "tsqueue":
-      return baseUri.with({ query: `topic=commands-set-tsqueue-tsqname` });
-
-    default:
-      return baseUri.with({ query: `topic=commands-set-${resourceType.toLowerCase()}` });
+  if (queryParam && fragment) {
```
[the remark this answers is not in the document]
almightyrush: Renamed it to anchors and soft removed the anchor links.

packages/vsce/__tests__/__unit__/docUrls.unit.test.ts
```diff
@@ -9,36 +9,24 @@
...
-  "pipeline",
-  "program",
-  "task",
-  "tcpipservice",
-  "transaction",
-  "tsqueue",
-  "urimap",
-  "webservice",
-];
+const resources = getMetas()
```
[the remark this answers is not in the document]
almightyrush: Done!

packages/vsce/src/utils/urlUtils.ts
```diff
@@ -26,17 +27,13 @@ export function generateDocumentationURL(resourceType?: string): Uri {
...
-      return baseUri.with({ query: `topic=commands-set-tsqueue-tsqname` });
-
-    default:
-      return baseUri.with({ query: `topic=commands-set-${resourceType.toLowerCase()}` });
+  if (queryParam) {
+    return baseUri.with({ query: `topic=${queryParam}` });
   }
+
+  // Fallback to default pattern
+  return baseUri;
```
davenice: In this situation we just take the user back to the default page of the help, right?

I'm inclined to think that this is not very helpful and that in this situation we'd be better not showing the option to open the documentation. 🤔 

I could be persuaded otherwise! (And we can leave it as it is for this PR)

packages/vsce/src/errors/CICSErrorHandler.ts
```diff
@@ -27,18 +26,18 @@ export function resourceNotFoundError(error?: ICICSExtensionError) {
...
-    const result = this.notifyErrorMessage({ errorMessage: msg, action: actions });
 
+    let message = msg;
     if (resourceType && !action) {
-      result.then(async (selection) => {
-        if (selection === URLConstants.OPEN_DOCUMENTATION) {
-          await openDocumentation(resourceType.trim().toLowerCase());
-        }
-      });
+      const docUrl = generateDocumentationURL(resourceType.trim().toLowerCase()).toString();
```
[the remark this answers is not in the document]
almightyrush: Thanks, I have externalised it.

PR #580, opened 2026-02-05
Update PR template to include an explicit issue link

davenice:
**What It Does**
Lots of our PRs have missed the link back to the issue. This is important information because it allows you to look at an issue and see what PRs contributed to resolving it. It also helps us close issues that are completed!

I'm also removing the changelog tick box because we have a check that verifies it's been updated.


enamkhan (approved):
LGTM!

PR #593, opened 2026-02-18
Displaying clear error and log messages for profile connection failure

### LINKED ISSUE #581: Error and log messages are not clear which profile they relate to
**Describe the bug**

With most error messages that we show the user, we're not clear about which profile the error has happened against.

Example of connecting to a CICSplex that is down:

Image

Or disabling DFHRPL:

Image

Similarly in our log messages you have to guess which profile they might relate to:

Image

This makes it hard to debug. We should notate which profile the error relates to. Perhaps:
> Failed to DISABLE LIBRARY DFHRPL on profile.name with API: PERFORM SET, RESP: 16 (INVREQ) and RESP2: 6. Please refer to the [IBM documentation](https://www.ibm.com/docs/en/cics-ts/6.x?topic%3Dcommands-set-library) for additional details

or

> Failed to send request on profile.name. Response details - URL: /CICSSystemManagement/CICSManagedRegion/CICSEX61/, Message: Failed to send an HTTP request.

harshithajahagirdar:
**What It Does**
Improves error messages and logs to show which profile we're working with.
Also logs request body when we make a PUT request.

**How to Test**

**Which Issue It Relates To**
Resolves https://github.com/zowe/cics-for-zowe-client/issues/581

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**


almightyrush:
> Please can I have validation that it's OK for us to change the SDK API like this, @almightyrush / @AndrewTwydell.

@davenice, The profile won't be available for sdk, as it is local for vsce and only few params are passed for calling the api, so the profile details can be fetched when doing the GET | PUT call within the vsce. 
Also the `ICMCIResponseErrorFeedBack` will contain the details from the cmci and it won't have any info related to the profile.

davenice:
I think profiles are available in the CLI as well; it's a Zowe-wide concept. I do agree that the profile doesn't fit all that well into CMCI feedback .... however every CMCI request from Zowe will have a profile associated.

I'll have another poke.

davenice:
OK - take 2. I realised I'd not pushed the right code when I made the last comment, so I deleted it.

I've removed the SDK changes and all is still good. I think they might have been an earlier iteration of the thinking on this.

AndrewTwydell (approved):
Looks good. Question around the optional nature of profileName..!

almightyrush (approved):
LGTM!

PR #595, opened 2026-02-23
Trigger comparison from RI and tree

Rajat-Sahrawat:
**What It Does**
User can trigger a comparison from a CICS resource in the tree and an open resource in the Resource Inspector

**How to Test**
Right click on a resource in a tree > Select "Compare to..."
or
Open a resource in Resource Inspector and select "Compare to..." from the menu option

**Which Issue It Relates To**

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**


davenice:
Hey Rajat,

We talked over this in the Zowe scrum today. In particular we talked about the tree behaviour - it would be good to bring this in line with VS Code's normal behaviour.

I know that before you made this PR, we weren't in line with VS Code, but let's not add the "Compare to..." to the tree.

We can either update the behaviour in this PR to be in line with VS Code, or we can create a new issue for it.

Here's a reminder of the VS Code behaviour:
https://github.com/user-attachments/assets/d435b975-f80d-46c3-8f61-66d75ca6b0da

Essentially you have:
`Select for Compare`
`Compare with Selected`  <<-- only if something has already been selected

I think this would address the usability concerns we had with multi-select and bring us in line with VS Code and Zowe Explorer.

PR #606, opened 2026-03-08
Implement generic reveal-in-tree functionality

### LINKED ISSUE #449: Inquire program does not work from Resource Inspector
**Describe the bug**

You can right-click on a transaction and choose "Inquire Program". This opens the programs node for the appropriate region and filters on the correct program.

We offer this action on the Resource Inspector as well.

This works as expected if you open the RI from the tree and therefore the programs node is showing. However if the programs node isn't showing; perhaps you open RI from the command palette or you collapse the region node of the tree, then inquire program does nothing.

BRSinha08:
When clicked inquire program or show library or inquire transaction  from Resource inspector or CICS Plex /tree , it must open the respective resource.

This is achieved by implementing a generic revealNodeInTree method that takes a resource and it's context (profile, CICSplex, region) and reveals it in the tree. This can be utilised by inquireProgram, inquireTransaction, etc, from the tree OR resource inspector.

**Review Checklist**
I certify that I have:
- [x] tested my changes

Resolves #449


packages/vsce/src/commands/inquireProgram.ts
davenice: Can this use the same revealResourceInTree API? then your code ends up looking like this:

- decide if we're in the region tree, resource inspector, or "all x tree"
- arrange all of the arguments correctly
- call revealResourceInTree

I think aiming to use the one function will mean that things can be more reliable because we use the same code path more (and so it gets tested for free!)

If this is hard to achieve or doesn't look good in practice we can discuss!

packages/vsce/src/commands/inquireProgram.ts
```diff
@@ -28,32 +29,42 @@ export function getInquireProgramCommand(tree: CICSTree, treeview: TreeView<any>
...
-      programTree = nodes[0]
-        .getParent()
-        .getParent()
-        .children.filter((child: CICSResourceContainerNode<IResource>) =>
-          child.resourceTypes.includes(ProgramMeta)
-        )[0] as CICSResourceContainerNode<IProgram>;
+    // Pull program(s) from transaction(s) to reveal and collate as criteria
+    const programNames = nodes
+      .map((n) => n.getContainedResource().resource.attributes.program)
+      .filter((name): name is string => !!name);
```
AndrewTwydell: What does this line do, and why do we need it?

packages/vsce/src/commands/inquireTransaction.ts
```diff
@@ -28,32 +30,49 @@ export function getInquireTransactionCommand(tree: CICSTree, treeview: TreeView<
...
-        .children.filter((child: CICSResourceContainerNode<IResource>) =>
-          child.resourceTypes.includes(TransactionMeta)
-        )[0] as CICSResourceContainerNode<ITransaction>;
+    // Verify it's a Task resource
+    const firstNode = nodes[0];
+    const containedResource = firstNode.getContainedResource();
+    if (containedResource.meta.resourceName !== CicsCmciConstants.CICS_CMCI_TASK) {
+      window.showErrorMessage(l10n.t("This command can only be used with Task resources"));
+      return;
     }
```
AndrewTwydell: This step should not be necessary as we can only run this from a Task. You can make this type-safe by adding the type annotation in the method 
```typescript
return commands.registerCommand("cics-extension-for-zowe.inquireTransaction", async (node: CICSResourceContainerNode<ITask>) => {
```

packages/vsce/src/commands/revealNodeInTree.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  if (criteria.length === 1) {
+    window.showInformationMessage(
+      l10n.t("{0} '{1}' revealed in the tree", resourceMeta.humanReadableNameSingular, criteria[0])
+    );
+  } else {
+    window.showInformationMessage(
+      l10n.t("Found {0} {1} and filter applied", criteria.length, resourceMeta.humanReadableNamePlural.toLowerCase())
+    );
+  }
```
AndrewTwydell: I don't think we need to show a pop-up message to the user that the action they asked for has happened; they should be able to see that it's happened!! 

Only need to show something when it has errored

packages/vsce/src/commands/revealNodeInTree.ts
```diff
@@ -0,0 +1,288 @@
...
+  let inspectorResources: any[] | undefined;
+  if (isFromInspector) {
+    const resourceInspector = ResourceInspectorViewProvider.getInstance(null, tree);
+    inspectorResources = resourceInspector?.getResources();
+  }
+  
+  return {
+    isFromInspector,
+    nodes,
+    inspectorResources,
```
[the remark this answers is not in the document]
BRSinha08: Fixed with new approach

packages/vsce/src/commands/inquireProgram.ts
AndrewTwydell: `const programNames = nodes.map((n) => n.getContainedResource().resource.attributes.program);` should be sufficient. 

We shouldn't have `null/undefined/empty program values` and will fail long before this step if we do!

Issue #575, opened 2026-02-03
Error when filtering plex resources on resource tree


JWaters02:
**Describe the bug**

I am trying to do a search for a program on this entire plex region

<img width="307" height="37" alt="Image" src="https://github.com/user-attachments/assets/eae0daa3-1047-4c74-b630-8f1e9856a605" />

when I click on search > select program > enter my filter, I get this error

<img width="548" height="272" alt="Image" src="https://github.com/user-attachments/assets/6b8bce8f-7dc4-4175-b171-52ce639ec8fb" />

and the "Loading resources" notification is stuck forever loading.

I set logging to trace but did not see any messages or errors output to logs.

I also tried restarting vscode to get a fresh state and got the issue on any of the plexes I tried.

**Expected and actual results**

Expected it to find the program in any of the regions in the plex.

**Describe your environment**

- Zowe Explorer version installed: 3.4.0
- Zowe Explorer for IBM CICS Transaction Server: 3.17.0
- Operating system and version: Win11H2

**Additional context**

davenice:
Thank you @JWaters02 ... interestingly I had wondered whether we should remove the ability to do this - applying the same filter across multiple regions at once. I had been thinking that searching across the whole CICSplex or individual regions might be sufficient.

I think we broke it a little while ago! Do you have an interesting use case for it or strong feeling about it? (I feel it should either work or not be there .... there are other ways to achieve it so I wouldn't feel bad removing it.)

davenice:
@JWaters02 - we have resolved this by removing the confusing filtering! Hopefully that works for you :-)

PR #610, opened 2026-03-17
Resolve confusing CICSplex filtering

harshithajahagirdar:
**What It Does**
1. Removes multiple options from region filter.
2. Will only filter regions.
3. Will resolve the errors that we are facing currently.

**How to Test**

**Which Issue It Relates To**

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**

For more information check the issue: https://github.com/zowe/cics-for-zowe-client/issues/575


packages/vsce/src/commands/getFilterPlexResources.ts
```diff
@@ -36,97 +34,15 @@ export function getFilterPlexResources(tree: CICSTree, treeview: TreeView<any>)
...
-                  }
-                }
-              }
-            }
-            return;
-          }
-        );
-      }
+    // Only filter regions
+    const resourceHistory = PersistentStorage.getSearchHistory(CicsCmciConstants.CICS_CMCI_MANAGED_REGION);
```
davenice: This bit of code caught my eye because we abstracted that all out into the meta files (If you search for PersistentStorage.getSearchHistory you'll see that now you've tidied it up the only calls are one here and all the others from the meta files).

This then got me thinking now that we've made region filters behave like all of the other filters, I wonder if we could just use filterResourceCommands to filter regions?

I'm not sure - it might not be feasible in this case - maybe just spend 15 mins trying to see whether it's an easy win - but I thought I'd mention it as the kind of suggestion that helps neaten things up.

davenice (changes requested):
This is looking pretty good I think.

One functional oddness - when you filter, the filter gets added to the tree node twice:

Image

The first one is not the right style, the second one is the correct style.

If you then refresh, the correct one disappears and the wrong one stays. Also we lose the region count:

Image

This behaviour works correctly for resources. Perhaps we need to lean on some of the resources node code better.

davenice:
@almightyrush / @AndrewTwydell - a quick re-review?

PR #628, opened 2026-03-30
Correct filtering when we have set a region in the zowe config file

harshithajahagirdar:
**What It Does**

1. If you have a region named in your zowe config
2. When you expand the cicsplex and then regions node
3. The regions node should show a filter set matching what was named in the zowe config
4. You should be able to set a new filter or remove the existing filter
5. If you close and open the regions node again i think it could revert to the original filter

**How to Test**

**Which Issue It Relates To**

**Review Checklist**
I certify that I have:
- [ ] tested my changes

**Additional Comments**


harshithajahagirdar:
Will create a new PR

Issue #18, opened 2024-02-01
Increase code-coverage for the CICS VSCE


zFernand0:

davenice:
We have been working further on this so I'm going to reopen it.

BRSinha08:
<img width="3396" height="1200" alt="Image" src="https://github.com/user-attachments/assets/7067d3b2-3781-44a8-b288-4d1cb3d919d2" />

the code coverage for main branch is 98.73% now

PR #629, opened 2026-04-01
Unit test for VSCE package

BRSinha08:
**Review Checklist**
I certify that I have:
- [x] tested my changes

Working towards https://github.com/zowe/cics-for-zowe-client/issues/18

### DIFF
--- a/packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
+++ b/packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
@@ -56,7 +56,7 @@ describe("actionResourceCommand", () => {
     } as imperative.IProfileLoaded;
 
     // Create mock tree - use object literal with proper typing instead of instantiating
-    const mockEventEmitter = new EventEmitter<any>();
+    const mockEventEmitter = new EventEmitter<CICSResourceContainerNode<IResource> | undefined>();
     mockEventEmitter.fire = jest.fn();
     
     mockTree = {
@@ -76,7 +76,7 @@ describe("actionResourceCommand", () => {
       getParent: jest.fn(),
       getConfigLocationPrompt: jest.fn(),
       getProfileIcon: jest.fn(),
-    } as unknown as CICSTree;
+    } as Partial<CICSTree> as CICSTree;
 
     // Create mock parent node
     mockParentNode = {
@@ -88,7 +88,7 @@ describe("actionResourceCommand", () => {
       getLabel: jest.fn(),
       regionName: "TESTREGION",
       cicsplexName: "TESTPLEX",
-    } as unknown as CICSResourceContainerNode<IResource>;
+    } as Partial<CICSResourceContainerNode<IResource>> as CICSResourceContainerNode<IResource>;
 
     // Create mock node with proper typing
     mockNode = {
@@ -109,7 +109,7 @@ describe("actionResourceCommand", () => {
         },
       }),
       getContainedResourceName: jest.fn().mockReturnValue("TESTPROG"),
-    } as unknown as CICSResourceContainerNode<IResource>;
+    } as Partial<CICSResourceContainerNode<IResource>> as CICSResourceContainerNode<IResource>;
 
     mockProgress = {
       report: jest.fn(),
@@ -171,7 +171,7 @@ describe("actionResourceCommand", () => {
       const mockNode2 = {
         ...mockNode,
         getContainedResourceName: jest.fn().mockReturnValue("TESTPROG2"),
-      } as unknown as CICSResourceContainerNode<IResource>;
+      } as Partial<CICSResourceContainerNode<IResource>> as CICSResourceContainerNode<IResource>;
 
       await actionTreeItem({
         action: "DISABLE",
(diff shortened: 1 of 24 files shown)


packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
```diff
@@ -0,0 +1,325 @@
...
+jest.mock("../../../src/errors/CICSErrorHandler");
+jest.mock("../../../src/utils/resourceUtils");
+jest.mock("../../../src/utils/treeUtils");
+
+describe("actionResourceCommand", () => {
+  let mockNode: any;
+  let mockTree: any;
+  let mockParentNode: any;
+  let mockProgress: any;
+  let mockToken: any;
```
AndrewTwydell: How possible is it to use actual types here (eg. `let mockTree: CICSTree`)? This will create more robust tests as they know the shape the variables should be. 

Then, if someone changes the CICSTree class for example, the unit tests will fail which is a good thing! Otherwise, the mock implementation will not accurately represent the class

AndrewTwydell: This seems to be the case in all the test cases

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
```diff
@@ -0,0 +1,325 @@
...
+        nodes: [mockNode, mockNode2],
+        tree: mockTree,
+      });
+
+      expect(mockTree.refresh).toHaveBeenCalledTimes(1);
+      expect(mockTree.refresh).toHaveBeenCalledWith(mockParentNode);
+    });
+
+    it("should handle different action types", async () => {
+      const actions = ["ENABLE", "DISABLE", "NEWCOPY", "PHASEIN"] as const;
```
AndrewTwydell: I've not seen this before! Why `as const` at the end?!

I would've thought this is the typing annotation, if needed at all!
```suggestion
      const actions: string[] = ["ENABLE", "DISABLE", "NEWCOPY", "PHASEIN"];
```

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
BRSinha08: fixed

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
BRSinha08: fixed

packages/vsce/__tests__/__unit__/utils/__mocks__/commonMocks.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export function createMockZoweAPI(profilesCache?: any) {
+  const cache = profilesCache || createMockProfilesCache();
+  return {
+    getExplorerExtenderApi: jest.fn().mockReturnValue({
+      getProfilesCache: jest.fn().mockReturnValue(cache),
+      initForZowe: jest.fn().mockResolvedValue(undefined),
+    }),
+  };
+}
```
AndrewTwydell: This method, as well as a lot of other stuff in this file, should already be covered by the mock of zowe-explorer-api in `packages/vsce/__tests__/__mocks__/@zowe/zowe-explorer-api.ts`.

That mocks the package and exposes loads of mocked methods you can set the return values for etc. We're using it in a lot of other places, wonder if you can also do this.

There may be gaps, which you can add, but probably not many!

packages/vsce/__tests__/__unit__/commands/compareResourceCommand.test.ts
```diff
@@ -0,0 +1,311 @@
...
+jest.mock("../../../src/commands/inspectResourceCommandUtils");
+
+describe("compareResourceCommand", () => {
+  let mockNode: any;
+  let mockContext: ExtensionContext;
+  let mockProfile: any;
+  let mockSession: any;
+  let mockResource: any;
+  let programMeta: any;
+
```
AndrewTwydell, line 36: Most of these test files use `any` as types. I think we discussed before that using their actual types (eg. CICSResourceContainerNode) would make these tests even **more** useful as they'll fail if interfaces change etc.

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
AndrewTwydell: There are still lots of cases where `any` is used in the tests. While this is easier, it'll make the tests even more useful

BRSinha08: used actuals as suggested

packages/vsce/__tests__/__unit__/utils/__mocks__/commonMocks.ts
BRSinha08: fixed as suggested

AndrewTwydell (approved):
Great thanks Bikash!

davenice (approved):

PR #639, opened 2026-04-16
fix(hyperlinks): accept a trailing slash in USS path pattern

### LINKED ISSUE #638: JAVA_HOME in JVM server is not hyperlinked
Lots of attributes now have hyperlinks to Zowe Explorer!

However JAVA_HOME in JVM server hasn't matched the regular expressions.

Image

Would be good to investigate and tweak the regular expression to catch whatever is different about it.

tmchow:
## Summary

Allow a single trailing slash in `USS_PATH_PATTERN` so directory values like `JAVA_HOME=/usr/lpp/java/J8.0_64/` (as reported by CICS JVM server) render as hyperlinks in the Resource Inspector instead of plain text. Addresses #638.

## Why

The existing regex in `packages/vsce/src/webviews/resource-inspector-panel/utils/hyperlinkUtils.tsx`:

```ts
const USS_PATH_PATTERN = /^\/[a-zA-Z0-9_\-.]+(\/[a-zA-Z0-9_\-.]+)*$/;
```

accepts `/u/user/file.txt` and `/opt/java` but rejects anything that ends with `/`, because each path segment after the leading slash requires at least one non-slash character to follow each slash. That's fine for file paths, but directories — which is exactly what JAVA_HOME is — conventionally end with a trailing slash on z/OS. The result: `/usr/lpp/java/J8.0_64/` flows through `renderHyperlinkableValue` unchanged and the user loses the click-through to Zowe Explorer that every other USS path gets.

The maintainer asked to "tweak the regular expression to catch whatever is different about it"; the only thing different is the trailing slash.

## Changes

- `hyperlinkUtils.tsx`: append `\/?` to the end of the pattern so exactly zero-or-one trailing slash is allowed. Updated the surrounding comment to explain why. Every other branch (consecutive slashes, invalid characters, empty paths, root-only `/`) is unchanged.

Quick sanity check (stays true after the change):

| Value | Before | After |
| --- | --- | --- |
| `/usr/lpp/java/J8.0_64` | ✅ | ✅ |
| `/usr/lpp/java/J8.0_64/` | ❌ | ✅ (new) |
| `/u/user/file.txt` | ✅ | ✅ |
| `/opt/java/` | ❌ | ✅ (new) |
| `/` | ❌ | ❌ |
| `//` | ❌ | ❌ |
| `/a//b` | ❌ | ❌ |

- `hyperlinkUtils.test.ts`: new `should return true for directory USS paths with a single trailing slash` case covering `/usr/lpp/java/J8.0_64/`, `/opt/java/`, `/u/cicsts/`, and `/a/b/`. Existing "should return false for empty or invalid paths" still holds — `/`, `//`, `/a//b` stay rejected.

## Testing

- `npx jest --config unit.jest_config.ts --testPathPatterns hyperlinkUtils` — 29 tests passed (25 existing + 4 new trailing-slash assertions).
- Verified the rejection cases (`/`, `//`, `/u/user//file`, `/a//b`) against the new regex directly via `node -e` to confirm they still fail.

Fixes #638

Signed-off-by: Trevin Chow <trevin@trevinchow.com>

This contribution was developed with AI assistance (Claude Code).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/packages/vsce/CHANGELOG.md
+++ b/packages/vsce/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the "cics-extension-for-zowe" extension will be documente
 ## Recent Changes
 
  - Enhancement: Adds hyperlinks to appropriate attributes in RI highlights. [#621](https://github.com/zowe/cics-for-zowe-client/issues/621)
+ - BugFix: Render USS paths with a trailing slash (e.g. `JAVA_HOME=/usr/lpp/java/J8.0_64/`) as hyperlinks in the Resource Inspector. [#638](https://github.com/zowe/cics-for-zowe-client/issues/638)
 
 ## `3.21.0`
 
--- a/packages/vsce/__tests__/__unit__/webviews/hyperlinkUtils.test.ts
+++ b/packages/vsce/__tests__/__unit__/webviews/hyperlinkUtils.test.ts
@@ -262,6 +262,15 @@ describe("hyperlinkUtils", () => {
       expect(isUssPathValue("/var/cics/region1/logs/messages.log")).toBe(true);
     });
 
+    test("should return true for directory USS paths with a single trailing slash", () => {
+      // JAVA_HOME values from CICS JVM servers often end with a trailing slash
+      // (e.g. /usr/lpp/java/J8.0_64/). See #638.
+      expect(isUssPathValue("/usr/lpp/java/J8.0_64/")).toBe(true);
+      expect(isUssPathValue("/opt/java/")).toBe(true);
+      expect(isUssPathValue("/u/cicsts/")).toBe(true);
+      expect(isUssPathValue("/a/b/")).toBe(true);
+    });
+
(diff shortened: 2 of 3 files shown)


davenice (changes requested):
This looks good to me.

Please could you update the changelog: https://github.com/tmchow/cics-for-zowe-client/blob/osc/638-uss-path-trailing-slash/packages/vsce/CHANGELOG.md to reference what you did and which issue it relates to.

If you rebase on main you'll find a recent changes section has been added that you can contribute to.

Thanks for taking the time to help out!

tmchow:
Thanks @davenice. Rebased onto main and added a `BugFix:` entry under the Recent Changes section referencing #638. Let me know if the wording needs adjusting.

davenice (approved):

AndrewTwydell (approved):

PR #643, opened 2026-04-28
New Copy programs nested under library dataset

### LINKED ISSUE #642: Error when NewCopying programs under a CICS library data set.
**Describe the bug**

When performing a newcopy on programs inside a library resource in the tree, I get this error. 

`Error running command cics-extension-for-zowe.newCopyProgram: Cannot destructure property 'errorMessage' of 'e.cicsExtensionError' as it is undefined.. This is likely caused by the extension that contributes cics-extension-for-zowe.newCopyProgram.`

Should newcopy programs as if they were under the `Programs` tree.

AndrewTwydell:
**What It Does**

When newcopying programs nested under a CICS library, as part of a CICS library dataset, an error occurs. This PR addresses this, parsing the correct set of args to the action allowing a newcopy on 1 or more programs under a library.

**How to Test**

NewCopy a program shown under a CICSLibrary. 

**Which Issue It Relates To**
Resolved #642 

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**

### DIFF
--- a/packages/vsce/src/commands/newCopyCommand.ts
+++ b/packages/vsce/src/commands/newCopyCommand.ts
@@ -30,15 +30,16 @@ export function getNewCopyCommand(tree: CICSTree, treeview: TreeView<any>) {
       return;
     }
 
-    if(clickedNode.contextValue?.includes("PARENT.CICSLibraryDatasetName")){
-      return actionTreeItem({ 
+    // Programs under library datasets require parent resource information for NEWCOPY
+    if (clickedNode.contextValue?.includes("PARENT.CICSLibraryDatasetName")){
+      await actionTreeItem({
         action: "NEWCOPY", 
         nodes, 
         tree, 
-        getParentResource: (node: CICSResourceContainerNode<IProgram>) => node.getContainedResource().resource.attributes  
+        getParentResource: (node: CICSResourceContainerNode<IProgram>) => node.getContainedResource().resource.attributes
       });
     }
 
-    return actionTreeItem({ action: "NEWCOPY", nodes, tree });
+    await actionTreeItem({ action: "NEWCOPY", nodes, tree });
   });
 }
--- a/packages/vsce/src/commands/newCopyCommand.ts
+++ b/packages/vsce/src/commands/newCopyCommand.ts
@@ -30,6 +30,15 @@ export function getNewCopyCommand(tree: CICSTree, treeview: TreeView<any>) {
       return;
     }
 
-    await actionTreeItem({ action: "NEWCOPY", nodes, tree });
+    if(clickedNode.contextValue?.includes("PARENT.CICSLibraryDatasetName")){
+      return actionTreeItem({ 
+        action: "NEWCOPY", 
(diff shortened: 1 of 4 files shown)


packages/vsce/src/commands/newCopyCommand.ts
```diff
@@ -30,6 +30,16 @@ export function getNewCopyCommand(tree: CICSTree, treeview: TreeView<any>) {
       return;
     }
 
+    // Programs under library datasets require parent resource information for NEWCOPY
+    if (clickedNode.contextValue?.includes("PARENT.CICSLibraryDatasetName")){
```
chinmdas, line 34: Using `includes("PARENT.CICSLibraryDatasetName")` can match the wrong thing if the string format changes or contains extra tags. 
for example, say `clickedNode.contextValue=PARENT.CICSLibraryDatasetNameSomeParentTag` condition will still be `true`.

what I think, this hard condition check would be better
```
const PARENT_LIBRARY_DATASET_CONTEXT = "PARENT.CICSLibraryDatasetName";

if (clickedNode.contextValue === PARENT_LIBRARY_DATASET_CONTEXT) {
```

This is just a suggestion, let me know your thoughts

chinmdas (changes requested):
Just a minor suggestion. Overall looks great to me..!!!!

packages/vsce/src/commands/newCopyCommand.ts
AndrewTwydell: The `includes` is kind of the point here, as the context value will have leading and trailing values. This is to check if the context value has that parent block within it. 

`CICSResource.CICSProgram.MYPROG1`
vs
`CICSResource.CICSProgram.MYPROG1.PARENT.CICSLibraryDatasetName`

AndrewTwydell: And there may be more stuff adding to the end of this in the future, so we can't use a `.endsWith(xx)`

packages/vsce/src/commands/newCopyCommand.ts
```diff
@@ -30,6 +30,16 @@ export function getNewCopyCommand(tree: CICSTree, treeview: TreeView<any>) {
...
+      await actionTreeItem({
+        action: "NEWCOPY", 
+        nodes, 
+        tree, 
+        getParentResource: (node: CICSResourceContainerNode<IProgram>) => node.getContainedResource().resource.attributes
```
BRSinha08: @AndrewTwydell just had a thought, if the node here is referring to parent, I guess the name should be parentNode instead of node, something like below for better clarity.

getParentResource: (parentNode: CICSResourceContainerNode<IProgram>) => parentNode.getContainedResource().resource.attributes

AndrewTwydell: Yup! Fixed 😄

chinmdas (approved):
Looks good now..!!!

BRSinha08 (approved):
LGTM!

PR #644, opened 2026-04-30
Enhance unit test coverage of VSCE package-2

BRSinha08:
Adding more unit tests and files to enhance the code coverage for VSCE package.

**Review Checklist**
I certify that I have:
- [x] tested my changes

Working towards https://github.com/zowe/cics-for-zowe-client/issues/18


packages/vsce/src/commands/showLogsCommand.ts
```diff
@@ -62,7 +62,7 @@ export async function getJobIdForRegion(selectedRegion: CICSRegionTree | CICSRes
 export function getShowRegionLogs(treeview: TreeView<any>) {
   return commands.registerCommand("cics-extension-for-zowe.showRegionLogs", async (node: CICSRegionTree | CICSResourceContainerNode<IResource>) => {
     // Handle both CICSRegionTree (from tree view) and CICSResourceContainerNode (from Resource Inspector)
-    const selectedRegion = node instanceof CICSRegionTree ? (node ?? treeview.selection[0]) : node;
+    const selectedRegion = node ?? treeview.selection[0];
```
AndrewTwydell: Are their any implications of changing this logic? It seems unrelated to the PR but perhaps not

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
-        resources: [],
+        resources,
       };
 
       expect(message.type).toBe("refresh");
-      expect(message.resources).toEqual([]);
+      if (message.type === "refresh") {
+        expect(message.resources).toHaveLength(1);
+        expect(message.resources[0].name).toBe("TESTPROG");
+      }
```
AndrewTwydell: We don't want conditionals in a unit test - we know the value

```suggestion
      expect(message.resources).toHaveLength(1);
      expect(message.resources[0].name).toBe("TESTPROG");
```

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+        type: "executeAction",
+        actionId: "newcopy",
+        resources,
+      };
+
+      expect(message.type).toBe("executeAction");
+      if (message.type === "executeAction") {
+        expect(message.actionId).toBe("newcopy");
+        expect(message.resources).toHaveLength(1);
+      }
```
AndrewTwydell: Same here about conditionals

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+        actionId: "refresh",
         resources: [],
       };
 
       expect(message.type).toBe("executeAction");
-      expect(message.actionId).toBe("testAction");
-      expect(message.resources).toEqual([]);
+      if (message.type === "executeAction") {
+        expect(message.resources).toEqual([]);
+      }
```
AndrewTwydell: And here

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+          regionName: "REGION1",
+          cicsplexName: undefined,
+        },
       };
 
       expect(message.type).toBe("showLogsForHyperlink");
       if (message.type === "showLogsForHyperlink") {
-        expect(message.resourceContext).toBeDefined();
+        expect(message.resourceContext.regionName).toBe("REGION1");
       }
```
AndrewTwydell: Remove this conditional too

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+          profile: {} as Partial<IResourceContext["profile"]> as IResourceContext["profile"],
+          session: {} as Partial<IResourceContext["session"]> as IResourceContext["session"],
+          regionName: "REGION1",
+          cicsplexName: undefined,
+        },
+        datasetName: "USER.DATASET",
       };
 
       expect(message.type).toBe("showDatasetForHyperlink");
       if (message.type === "showDatasetForHyperlink") {
```
AndrewTwydell: And this

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+          profile: {} as Partial<IResourceContext["profile"]> as IResourceContext["profile"],
+          session: {} as Partial<IResourceContext["session"]> as IResourceContext["session"],
+          regionName: "REGION1",
+          cicsplexName: undefined,
+        },
+        ussPath: "/u/user/file.txt",
       };
 
       expect(message.type).toBe("showUssFileForHyperlink");
       if (message.type === "showUssFileForHyperlink") {
```
AndrewTwydell: And this

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
```diff
@@ -9,117 +9,229 @@
...
+    });
+
+    it("should handle multiple resources in refresh message", () => {
+      const resources = [createMockResource(), createMockResource()];
+      const message: WebviewToExtensionMessage = {
+        type: "refresh",
+        resources,
+      };
+
+      if (message.type === "refresh") {
```
AndrewTwydell: And this

packages/vsce/src/commands/showLogsCommand.ts
AndrewTwydell: Thinking about this more, I don't think we should be changing the behaviour here. 

If it's a bug, we can create a new PR to solve the bug this is causing - it's not a change to be made in a unit test improvement PR

packages/vsce/__tests__/__unit__/utils/treeUtils.unit.test.ts
AndrewTwydell, line 1: Lets remove this file change - it's a missing newline at the end of the file

packages/vsce/__tests__/__unit__/utils/filterUtils.unit.test.ts
```diff
@@ -39,7 +40,7 @@ const createMockQuickPick = (value: string, selectedItems: any[] = []) => {
 const setupSingleQuickPick = (value: string, resolvedValue?: { label: string; description?: string }) => {
   const selectedItems = resolvedValue ? [resolvedValue] : [];
   const mockQuickPick = createMockQuickPick(value, selectedItems);
-  jest.spyOn(Gui, "createQuickPick").mockReturnValue(mockQuickPick as any);
+  jest.spyOn(Gui, "createQuickPick").mockReturnValue(mockQuickPick as Partial<QuickPick<QuickPickItem>> as QuickPick<QuickPickItem>);
```
AndrewTwydell, line 43: I don't particularly like all these `as Partial<xx> as Type<xx>` castings. There seem to be **loads** of them but they feel rather unmaintainable. Something to consider going forward!

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
```diff
@@ -63,9 +63,9 @@ describe("actionResourceCommand", () => {
       refresh: jest.fn(),
       _onDidChangeTreeData: mockEventEmitter,
       loadedProfiles: [],
-      getLoadedProfiles: jest.fn().mockReturnValue([]),
+      getLoadedProfiles: jest.fn(() => []),
```
AndrewTwydell: What's the difference here?

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
```diff
@@ -81,8 +81,8 @@ describe("actionResourceCommand", () => {
     // Create mock parent node
     mockParentNode = {
       refresh: jest.fn(),
-      getProfileName: jest.fn().mockReturnValue("testProfile"),
-      getProfile: jest.fn().mockReturnValue(mockProfile),
+      getProfileName: jest.fn(() => "testProfile"),
+      getProfile: jest.fn(() => mockProfile),
```
AndrewTwydell: And here, what's the difference?

packages/vsce/__tests__/__unit__/utils/commandUtils.unit.test.ts
```diff
@@ -21,7 +23,7 @@ jest.spyOn(ZoweVsCodeExtension, "getZoweExplorerApi").mockReturnValue({
   getJesApi: getJesApiMock,
   getMvsApi: getMvsApiMock,
   getUssApi: getUssApiMock,
-} as any);
+} as Partial<ReturnType<typeof ZoweVsCodeExtension.getZoweExplorerApi>> as ReturnType<typeof ZoweVsCodeExtension.getZoweExplorerApi>);
```
AndrewTwydell, line 26: This is tricky and totally unmaintainable. I don't think this adds any value so perhaps can be `any` in this case.

packages/vsce/__tests__/__unit__/commands/showLogsCommand.unit.test.ts
AndrewTwydell, line 1: There is a `showLogsCommand.test.ts` and `showLogsCommand.unit.test.ts`. What is the difference? Why do we need both? Is there overlap?

packages/vsce/__tests__/__unit__/commands/showLogsCommand.unit.test.ts
```diff
@@ -43,20 +100,43 @@ describe("Test suite for findRelatedZosProfiles", () => {
...
+        if (name.startsWith("host4")) return Promise.resolve({ name: "host4", profile: {}, type: "base" });
+        return Promise.reject(new Error("No base profile"));
+      }),
+    };
+    jest.spyOn(ProfileManagement, "getProfilesCache").mockReturnValue(mockProfileCache as any);
+  });
+
   it("Profile with common base finds z/osmf", async () => {
-    const profile = await findRelatedZosProfiles(createProfile("host1.mycics", "cics", "h1", "user"), zosProfiles);
+    const { findRelatedZosProfiles: mockFindRelatedZosProfiles } = require("../../../src/utils/commandUtils");
```
AndrewTwydell: We don't do imports mid-file anywhere else - why is this?

We tend to import things at the top, and create spies or mocks on their methods.

AndrewTwydell (changes requested):

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
BRSinha08: mockReturnValue([]): Returns the same array instance on every call
jest.fn(() => []): Returns a new array instance on every call.

jest.fn(() => []) is generally safer as it prevents unintended mutations from affecting subsequent calls, though for an empty array that's typically not modified, both approaches would work similarly in most test scenarios.

So I will revert it.

packages/vsce/__tests__/__unit__/commands/actionResourceCommand.test.ts
BRSinha08: here also, there will be not much difference, it's just taking the concise implementation without adding an extra method.
mockReturnValue() is more explicit and chain-able with other mock methods.

I will revert this.

packages/vsce/__tests__/__unit__/utils/filterUtils.unit.test.ts
BRSinha08: This approach is important for the test to focus on the behaviour being tested without implementing entire interface.(QuickPick).

In unit tests, you only need to mock the properties/methods that are actually used
Creating a complete mock with all QuickPick properties would be verbose and unnecessary
The test only interacts with the mocked properties, so the incomplete implementation is safe.

So, I think usage of partial is okay and , it can be maintained , as it makes all properties optional, allowing the incomplete mock to satisfy TypeScript and then casts to QuickPick<QuickPickItem> ------>tells TypeScript to treat it as the full interface for the test.

Please let me know, if it is still not worth, I will remove from all places.

packages/vsce/src/commands/showLogsCommand.ts
BRSinha08: reverted

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/webviews/common/messages.test.ts
BRSinha08: fixed as suggested

packages/vsce/__tests__/__unit__/commands/showLogsCommand.unit.test.ts
BRSinha08: fixed as suggested, kept showLogsCommand.test.ts

packages/vsce/__tests__/__unit__/commands/showLogsCommand.unit.test.ts
BRSinha08: fixed as suggested.

packages/vsce/__tests__/__unit__/utils/commandUtils.unit.test.ts
BRSinha08: same as other partial comment.

AndrewTwydell (approved):
Thanks! 😄

davenice (approved):

PR #659, opened 2026-06-11
Enhance region node with persistent filter, visual indicator, and config region default.

harshithajahagirdar:
1. The regions node now displays a filter indicator showing the active filter.
2. Users can see that a filter is in effect
3. Users can apply a new filter or clear the existing filter
4. If you close and open the Regions node again, the last applied filter persists during the session
5. The config region appears as the first option when opening the filter dialog.

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**


BRSinha08 (changes requested):
write the missing test cases

davenice:
An item to revisit in this PR currently:

When you set a region group in the config file, this both shows the region group applied at the CICSplex level (which is probably correct) and also attempts to filter the regions by the text of the region group name - this will not work because the region group will have a name that is different to the regions.

packages/vsce/src/utils/filterUtils.ts
```diff
@@ -25,9 +25,18 @@ export class FilterDescriptor implements QuickPickItem {
   }
 }
 
-export const buildQuickPick = (resName: string, history: string[]) => {
+export const buildQuickPick = (resName: string, history: string[], profileRegionName?: string) => {
```
davenice: This is a bit odd - it's a generic quick pick builder and we have a reference to a profileRegionName here.

The code does work (because resource quick pickers don't pass in a profileRegionName) but it's not very clean design.

What if we allow the caller to pass in objects instead of strings and a profileRegionName? The object could just be a `QuickPickItem` - so we could pass in `QuickPickItem[]` instead of `string[]` and convert them earlier in the code.

packages/vsce/src/utils/filterUtils.ts
```diff
@@ -25,9 +25,18 @@ export class FilterDescriptor implements QuickPickItem {
...
   const quickpick = Gui.createQuickPick();
-  quickpick.items = history.map((loadedFilter) => ({ label: loadedFilter }));
+  quickpick.items = history.map((loadedFilter) => {
+    // Add description to any item that matches the profile region name
+    if (profileRegionName && loadedFilter === profileRegionName) {
+      return {
+        label: loadedFilter,
+        description: l10n.t("Zowe CICS profile"),
```
davenice: Making that previous change also moves this "Zowe CICS profile" string out to getFilterPlexResources which is more where that string fits, rather than in this generic method.

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -77,143 +95,132 @@ export class CICSRegionsContainer extends TreeItem {
...
         return;
       }
-      
+
       const { regions, apiResponse } = regionInfo;
-      
-      // Handle errors if present (warnings for partial results)
       if (apiResponse) {
-        CICSErrorHandler.handleErrorIfPresent(apiResponse, CicsCmciConstants.DOC_RESOURCE_TYPE_GET, parentPlex.getProfile().name);
+        CICSErrorHandler.handleErrorIfPresent(apiResponse, "get", parentPlex.getProfile().name);
```
davenice: I think this has undone Bikash's constant use? Is that intentional?

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -77,143 +95,132 @@ export class CICSRegionsContainer extends TreeItem {
...
+        this.updateDescription();
       }
+    } catch (error) {
+      this.children = [];
+      CICSErrorHandler.handleCMCIRestError(
+        new CICSExtensionError({
+          baseError: error as Error | Record<string, unknown>,
+          profileName: this.parent.getProfile().name,
+          resourceName: "regions",
+          errorMessage: l10n.t("Failed to load regions in plex"),
```
davenice: Please can we call it CICSplex not plex? Generally a plex is not a thing, a CICSplex is a thing.

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -77,143 +95,132 @@ export class CICSRegionsContainer extends TreeItem {
...
         resourceName: CicsCmciConstants.CICS_CMCI_MANAGED_REGION,
         cicsPlex: plexProfile.profile.cicsPlex,
         regionName: plexProfile.profile.regionName,
       });
-      
-      // Handle errors if present
-      CICSErrorHandler.handleErrorIfPresent(apiResponse, CicsCmciConstants.DOC_RESOURCE_TYPE_GET, plexProfile.name);
-      
-      const regionsArray = toArray(apiResponse.response.records.cicsmanagedregion);
+      CICSErrorHandler.handleErrorIfPresent(regionsObtained, "get", plexProfile.name);
```
davenice: constant is lost!

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -77,143 +95,132 @@ export class CICSRegionsContainer extends TreeItem {
...
           }
         } catch (error) {
-          const wrappedError = new CICSExtensionError({
-            baseError: error as Error | Record<string, unknown>,
-            profileName: this.parent.getProfile().name,
-          });
-          await CICSErrorHandler.handleCMCIRestError(wrappedError);
           this.children = [];
-          // Set visual indicator for failed load
-          this.collapsibleState = TreeItemCollapsibleState.Collapsed;
```
davenice, line 87: There are a couple of places where we've lost this collapsing the tree back again.

Is that OK? I'm not a fan of the "Load Failed" message ... we get a notification anyway. So I'd be happy to lose that bit as long as the tree operates correctly if there's an error!

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -11,50 +11,66 @@
...
+   * Filters regions based on a pattern (supports wildcards and comma-separated patterns)
+   * @param pattern - Filter pattern (e.g., "CICS*", "CICS1,TEST*", or "*" for all)
+   */
+  public async filterRegions(pattern: string) {
     this.activeFilter = pattern;
+    // Save the filter to the parent plex so it persists across tree rebuilds
+    this.parent.saveRegionFilter(pattern);
     this.updateLabelAndContext();
-    await PersistentStorage.setCriteria(this.buildFilterStorageKey(), pattern === "*" ? undefined : pattern);
+    this.isRefreshing = true;
```
davenice: Could you help me understand why we need this? These kind of state flags are difficult to manage and get right.

It looks like we never use it.

packages/vsce/src/commands/clearPlexFilterCommand.ts
```diff
@@ -20,6 +20,8 @@ import type { CICSTree } from "../trees/CICSTree";
  */
 export function getClearPlexFilterCommand(tree: CICSTree) {
   return commands.registerCommand("cics-extension-for-zowe.clearPlexFilter", async (node: CICSRegionsContainer) => {
-    node.filterRegions("*", tree);
+    node.clearFilter();
+    await node.filterRegions(node.activeFilter);
```
davenice: We do too much work here.

node.clearFilter() calls filterRegions then immediately after we call node.filterRegions(...)

filterRegions does I/O so is something we don't want to do more than we need to.

packages/vsce/src/trees/CICSRegionsContainer.ts
```diff
@@ -11,50 +11,66 @@
...
-    const savedFilter = PersistentStorage.getCriteria(this.buildFilterStorageKey());
-    this.activeFilter = savedFilter || "*";
+    const profile = parent.getProfile().profile;
+    // For region groups
+    if (parent.getGroupName()) {
+      this.originalConfigFilter = "*";
+      this.activeFilter = "*";
+    } else {
+      this.originalConfigFilter = profile.cicsPlex && profile.regionName ? profile.regionName : "*";
+      this.activeFilter = savedFilter || this.originalConfigFilter;
```
davenice: This line here is the only place that originalConfigFilter is used. I think we can simplify and remove the variable.

packages/vsce/src/trees/CICSPlexTree.ts
```diff
@@ -184,11 +185,21 @@ export class CICSPlexTree extends TreeItem {
...
     this.regionsContainer = regionContainer;
   }
 
+  public saveRegionFilter(filter: string) {
+    this.savedRegionFilter = filter;
+    // Save to session tree so it persists across plex tree rebuilds
+    this.parent.saveRegionFilterForPlex(this.getPlexName(), filter);
+  }
+
+  public getSavedRegionFilter(): string | undefined {
```
davenice: I'm not sure we need this method - it's only used by tests that I can see. Could we just access the field directly?

davenice (changes requested):
I've tried to go through as thoroughly as I can. Add your changes as a new commit - it will be easier to see what you've changed like that. We can squash as it goes in (or you can squash once I've re-reviewed)

Thanks - I think it's broadly good, it's mainly tidying/simplifying.

packages/vsce/src/trees/CICSPlexTree.ts
harshithajahagirdar: Removed the method. savedRegionFilter is a public field, tests and any future callers access it directly, no getter needed.

packages/vsce/src/trees/CICSRegionsContainer.ts
harshithajahagirdar: Done

packages/vsce/src/commands/clearPlexFilterCommand.ts
harshithajahagirdar: Done, it only calls one now

packages/vsce/src/utils/filterUtils.ts
harshithajahagirdar: This sounds better. Implemented

packages/vsce/src/trees/CICSRegionsContainer.ts
harshithajahagirdar: This was unintentional— it's corrected

packages/vsce/src/trees/CICSRegionsContainer.ts
harshithajahagirdar: Done.

packages/vsce/src/trees/CICSRegionsContainer.ts
harshithajahagirdar: All three error paths now collapse the tree back and reset the icon.

packages/vsce/src/trees/CICSRegionsContainer.ts
harshithajahagirdar: You are right, its never used. Removed

packages/vsce/src/utils/patternUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Converts a filter pattern with wildcards into a regular expression
+ * Supports comma-separated patterns and wildcards (*)
+ *
+ * @param pattern - Filter pattern (e.g., "CICS*", "CICS1,TEST*", or "*")
+ * @returns Regular expression string for matching
+ *
+ * @example
+ * patternToRegex("CICS*") // Returns "(^CICS(.*))"
+ * patternToRegex("CICS*,TEST*") // Returns "(^CICS(.*))|(^TEST(.*))"
```
BRSinha08: these examples are not needed, you can remove it

BRSinha08 (changes requested):

packages/vsce/src/utils/filterUtils.ts
```diff
@@ -122,29 +122,29 @@ export function inputMatchesChoice(userInput: string, choiceLabel: string, caseS
   return normalizedInput === normalizedChoice;
 }
 
-export async function getPatternFromFilter(resourceName: string, resourceHistory: string[], filterCaseSensitive: boolean = false) {
-  const quickpick = buildQuickPick(resourceName, resourceHistory);
-  
+export async function getPatternFromFilter(resourceName: string, resourceItems: QuickPickItem[], filterCaseSensitive: boolean = false) {
```
BRSinha08: resourceItems can be historyItems , what do you think?

packages/vsce/src/utils/patternUtils.ts
harshithajahagirdar: Done — removed the @example blocks

packages/vsce/src/utils/filterUtils.ts
harshithajahagirdar: Agreed — renamed the parameter to historyItems

BRSinha08 (changes requested):
The Codecov should be green, there are missing tests, Please verify and if needed write it down.

davenice (approved):
Looks good to me I think. Thanks for all the work getting it here :-)

BRSinha08 (approved):
LGTM!

PR #664, opened 2026-06-23
Library and Local file(Enable & Disable) 

BRSinha08:
**What It Does**

- Introduces a comprehensive code generation system for CICS SDK resources using a resource-focused specification approach
- Adds a `codegen/` directory with TypeScript-based generators that create SDK code, unit tests, and related files from JSON specifications
- Implements a resource-specification-driven architecture where CICS resources (LocalFile, Library) are defined declaratively with their supported actions (CLOSE, OPEN, ENABLE, DISABLE)
- Generates SDK layer code including:
  - Resource action functions (e.g., `closeLocalFile`, `openLocalFile`, `enableLibrary`, `disableLibrary`)
  - Parameter interfaces (`ILocalFileParms`, `ILibraryParms`)
  - Constants and type definitions
  - Utility functions for resource actions
  - Unit tests for all generated functions
- Refactors existing LocalFile implementation to use the new generated code structure
- Adds new Library resource with ENABLE and DISABLE actions
- Introduces a CI check script to verify generated files are up-to-date
- Uses Handlebars templates for code generation with proper separation of concerns

**How to Test**

1. Test the code generator:cd codegen
    npm install
    npm run generate

2. Test the CI check:
     cd codegen
     npm run check:generated

3. Test generated SDK functions:
Run the generated unit tests: npm test in the SDK package
Verify all LocalFile actions (close, open, enable, disable) work correctly
Verify new Library actions (enable, disable) work correctly

**Which Issue It Relates To**
Working towards improved code maintainability and scalability for CICS resource management

**Review Checklist**
I certify that I have:
- [x] tested my changes

**Additional Comments**

**Architecture Highlights:**

1.Resource-Focused Specification: The resourceSpecification.json uses a declarative approach where resources are defined with their identifiers, actions, and options. Shared actions and options can be referenced across multiple resources, promoting DRY principles.

4.Template-Based Generation: Handlebars templates in codegen/templates/ separate presentation from logic, making it easy to modify generated code structure without touching the generator logic.

5.Type Safety: Full TypeScript support throughout with comprehensive interfaces for specifications and derived types.

6.Generated File Markers: All generated files include clear warning headers indicating they should not be manually edited.

7.Validation Layer: The generator includes parameter validation, length checks, and allowable value validation based on the specification.

8.CI Integration: The check-generated.ts script ensures generated files stay in sync with specifications in CI/CD pipelines.

9.Extensibility: Adding new resources or actions is as simple as updating the JSON specification and running the generator.

**Breaking Changes:**

1.Constants renamed for consistency (e.g., CICS_LOCAL_FILE_BUSY_VALUES → CICS_BUSY_VALUES)
2.Some internal SDK file structures reorganized (exports remain compatible)
3.Generated files now include warning headers

**Future Enhancements:**

1.Add more CICS resources (Programs, Transactions, etc.) to the specification
2.Generate CLI command definitions from the specification
3.Generate VSCE command handlers from the specification
4.Add integration tests for generated code


codegen/resourceSpecification.json
```diff
@@ -35,6 +35,39 @@
...
+          "identifier": {
+            "name": "DISABLE",
+            "aliases": ["dis"],
+            "group": "disable",
+            "description": "Disable a resource in CICS",
+            "verb": "disabling",
+            "verbPastTense": "disabled"
+          },
+          "options": []
+        }
```
AndrewTwydell: Because these are already defined in the `actions` block, can't you juse use their IDs?

```suggestion
        "ENABLE", "DISABLE"
```

packages/sdk/src/doc/ILibraryParms.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import type { IResourceParms } from "./IResourceParms";
+
+/**
+ * Interface for CICS library parameters
+ * @export
+ * @interface ILibraryParms
+ * @extends {IResourceParms}
+ */
+export interface ILibraryParms extends IResourceParms {
```
AndrewTwydell, line 27: I think it would be acceptable to ignore these lint errors when generating code. If it means we have a predictable interface structure for each resource, even if they're the same as the parent

AndrewTwydell (commented):
Looking good!

codegen/generate.ts
```diff
@@ -505,7 +543,19 @@ export class ResourceGenerator {
...
+    if (fs.existsSync(outputPath)) {
+      const existingContent = fs.readFileSync(outputPath, "utf-8");
+      if (existingContent === output) {
+        shouldWrite = false;
+      }
+    }
+    
+    if (shouldWrite) {
+      fs.writeFileSync(outputPath, output, "utf-8");
+    }
```
AndrewTwydell: What's the reasoning here? 

In my mind, ALWAYS doing a read and then sometimes doing a write is no more efficient than just ALWAYS doing a write. 

It also adds another layer of uncertainty of "did we write changes, did we not", with this logic when the output is always going to be the same. 

I'd be inclined to just always write. No behaviour will change, and we can be even more confident.

codegen/templates/sdk/parms.interface.hbs
```diff
@@ -0,0 +1,40 @@
...
+   * The busy condition option for closing the {{humanNameLower}}
+   * Valid values: "WAIT", "NOWAIT", "FORCE" (case-insensitive)
+   * Value will be automatically converted to uppercase
+   * @type {string}
+   * @memberof {{parmsInterface}}
+   */
+  busy?: string;
+{{else}}
+  // No additional parameters needed for {{humanName}} resource
+{{/if}}
```
AndrewTwydell: This is very specific to a couple of resource types - it'd be like having a resource template including conditions hardcoded for JVM options. 

I feel doing the same thing as elsewhere, where options are somewhat passed in. Is that possible?

BRSinha08: The busy parameter is now being generated from the options data rather than being hardcoded

codegen/generate.ts
BRSinha08: Earlier, it was adding more code unnecessarily, then I thought to add conditions, but if we just write it always, it worked , but going forward, I will verify in other PRs, whether it's working always or not.

AndrewTwydell (approved):

PR #693, opened 2026-07-03
program and uri map enable and disable

BRSinha08:
**What It Does**
Previously, SDK resource files (LocalFile.ts, Program.ts, URIMap.ts, Library.ts), their parameter interfaces, constants, and unit tests were written and maintained manually — leading to inconsistency between resources and duplication risk as the project grows. This PR introduces a complete Handlebars-based code generation system (codegen/) that becomes the single source of truth for all CICS resource SDK code. It also expands SDK coverage by adding enable/disable support for Library, Program, and URIMap, and wires the existing LocalFile enable/disable/open/close operations through the new performAction generic utility. A CI check script ensures generated files are never left out of sync with the specification.

✅ Fully implemented: Codegen system with resourceSpecification.json, Handlebars templates for resource files, parameter interfaces, constants, utils, index barrels, and unit tests
✅ Fully implemented: Generic performAction utility in ResourceActions.ts (generated) replacing per-resource CMCI boilerplate
✅ Fully implemented: SDK enable/disable for Library, Program, and URIMap; open/close/enable/disable for LocalFile
✅ Fully implemented: JSON Schema validation for the resource specification (resourceSpecification.schema.json)
✅ Fully implemented: CI guard script (check-generated.ts) — fails if committed generated files are stale
✅ Fully implemented: Parms interfaces (ILocalFileParms, IProgramParms, IURIMapParms) regenerated under codegen control, now extending the shared IResourceParms base
✅ Fully implemented: BUSY parameter modelled as a shared, reusable option constant; CICS_BUSY_VALUES replaces the per-resource CICS_LOCAL_FILE_BUSY_VALUES

**How to Test**
cd codegen && npm run generate   
git diff --name-only -- packages/sdk   

**Review Checklist**
I certify that I have:
- [x] tested my changes


codegen/templates/sdk/constants.hbs
```diff
@@ -58,7 +58,7 @@ export const CicsCmciConstants = {
   /**
    * Alias for CICS_LIBRARY_RESOURCE (used by codegen)
    */
-  CICS_CMCI_LIBRARY: "CICSLibrary",
+  CICS_CMCI_LIBRARY: "CICSLIBRARY",
```
AndrewTwydell: I'm wondering if all these constants need to be hardcoded in the template. Surely they'd be generated from the json spec? 

I'm a bit confused what they're doing, but you've got these hardcoded 
`CICS_LIBRARY_RESOURCE: "CICSLibrary"`
`CICS_CMCI_LIBRARY: "CICSLIBRARY"`

And then lower down you're generating them for everything but Library, LocalFile, UriMap.

codegen/generate.ts
```diff
@@ -350,14 +353,26 @@ export class ResourceGenerator {
...
+    // Special case for URIMap to use URI_MAP instead of U_R_I_MAP
+    let resourceTypeUpper: string;
+    if (sdkFileName === "URIMap") {
+      resourceTypeUpper = "URI_MAP";
+    } else {
+      resourceTypeUpper = nameWithoutCICS
+        .replace(/([A-Z])/g, "_$1")
+        .toUpperCase()
+        .replace(/^_/, "");
+    }
```
AndrewTwydell: Could be a good reason to store this in spec json in some form? We have different cases, perhaps need a snake_case key

codegen/generate.ts
```diff
@@ -350,14 +353,26 @@ export class ResourceGenerator {
...
+    
+    // Special case for URIMap to use existing CICS_URIMAP constant
+    const sdkResourceType = sdkFileName === "URIMap" ? "CICS_URIMAP" : `CICS_CMCI_${resourceTypeUpper}`;
+
```
AndrewTwydell: This is unexpected to me. I thought we were generating constants, so we can call it what we want?

codegen/resourceSpecification.json
```diff
@@ -11,7 +11,17 @@
...
+          "identifier": {
+            "name": "DISABLE",
+            "aliases": ["dis"],
+            "group": "disable",
+            "description": "Disable a resource in CICS",
+            "verb": "disabling",
+            "verbPastTense": "disabled"
+          },
+          "options": ["BUSY"]
+        },
```
AndrewTwydell, line 24: Is this different from the other Disable actions?

packages/sdk/__tests__/__unit__/disable/Disable.urimap.unit.test.ts
```diff
@@ -87,22 +107,20 @@ describe("CMCI - Disable urimap", () => {
         CicsCmciConstants.CICS_URIMAP +
         "/" +
         region +
-        `?CRITERIA=(NAME%3D${disableParms.name})`;
+        `?CRITERIA=(NAME%3D${ disableParms.name})`;
```
AndrewTwydell: Trivial but an annoying space there!

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -58,7 +58,7 @@ export const CicsCmciConstants = {
   /**
    * Alias for CICS_LIBRARY_RESOURCE (used by codegen)
    */
-  CICS_CMCI_LIBRARY: "CICSLibrary",
+  CICS_CMCI_LIBRARY: "CICSLIBRARY",
```
AndrewTwydell: Yeah this doesn't seem to fit the rest of the patterns 🤔

BRSinha08:
> @BRSinha08 , I have noticed that some of the
> 
> * **Unit tests** in cli (`packages/cli/__tests__/__unit__/enable/urimap/Urimap.handler.unit.test.ts` , `packages/cli/__tests__/__unit__/disable/urimap/Urimap.handler.unit.test.ts)`
> * **System tests** in sdk (`packages/sdk/__tests__/__system__/enable/Enable.urimap.system.test.ts`, `packages/sdk/__tests__/__system__/disable/Disable.urimap.system.test.ts`)
> 
> is still referencing the older `enableUrimap` and `disableUrimap`
> 
> Any reasons why you have not updated them OR added new tests cases to test your implementation?

> * **Unit tests** in cli:--> The first one mentioned above is from CLI package, which is not handled in this PR.
> * **System tests** in sdk:-->The second one is system test, so I am not generating system tests using codegen currently.

codegen/resourceSpecification.json
BRSinha08: yes it has busy parameters, others don't have

packages/sdk/__tests__/__unit__/disable/Disable.program.unit.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    it("should be able to disable a program", async () => {
+      endPoint =
+        "/" +
+        CicsCmciConstants.CICS_SYSTEM_MANAGEMENT +
+        "/" +
+        CicsCmciConstants.CICS_CMCI_PROGRAM +
+        "/" +
+        region +
+        `?CRITERIA=(program%3D${disableParms.name})`;
```
AndrewTwydell: I notice other resources have been asserting all Uppercase here - not a blocker, but something to consider for consistency. 

program= vs PROGRAM=
file= vs FILE=

packages/sdk/__tests__/__unit__/disable/Disable.urimap.unit.test.ts
```diff
@@ -90,19 +110,17 @@ describe("CMCI - Disable urimap", () => {
...
         request: {
-          update: {
-            attributes: {
-              $: {
-                ENABLESTATUS: "DISABLED",
-              },
+          action: {
+            $: {
+              name: "DISABLE",
             },
```
AndrewTwydell: This changing suggests we're now sending a different request body to CMCI - is this the case? 

Have you verified that we're generating hte correct structure? It concerns me that it's different

packages/sdk/__tests__/__unit__/enable/Enable.urimap.unit.test.ts
```diff
@@ -90,19 +110,17 @@ describe("CMCI - enable urimap", () => {
...
         request: {
-          update: {
-            attributes: {
-              $: {
-                ENABLESTATUS: "ENABLED",
-              },
+          action: {
+            $: {
+              name: "ENABLE",
             },
```
AndrewTwydell: Same here - were these tests here before, or are they new to the feature branch which you're now updating

packages/sdk/src/constants/CicsCmci.constants.ts
```diff
@@ -201,6 +191,21 @@ export const CicsCmciConstants = {
...
+  /**
+   * The CICS CMCI urimap
+   */
+  CICS_URIMAP: "CICSURIMap",
+
+  /**
+   * The CICS CMCI library
+   */
+  CICS_CMCI_LIBRARY: "CICSLibrary",
+
```
AndrewTwydell, line 208: We've got 2 prefixes here, `CICS_CMCI_` and `CICS_` - why is there a difference?

codegen/templates/sdk/constants.hbs
```diff
@@ -200,6 +190,15 @@ export const CicsCmciConstants = {
    * Valid BUSY parameter values for resource operations
    */
   CICS_BUSY_VALUES: ["WAIT", "NOWAIT", "FORCE"],
+{{#each resources}}
+{{#unless (eq sdkFileName "LocalFile")}}
```
davenice, line 194: Can you explain why LocalFile can't follow the same pattern? Is it because of something pre-existing that we can't break?

packages/sdk/__tests__/__unit__/disable/Disable.library.unit.test.ts
```diff
@@ -116,62 +97,23 @@ describe("CMCI - Disable library", () => {
...
+    it("should be able to disable a library", async () => {
       endPoint =
         "/" +
         CicsCmciConstants.CICS_SYSTEM_MANAGEMENT +
         "/" +
         CicsCmciConstants.CICS_CMCI_LIBRARY +
         "/" +
         region +
-        `?CRITERIA=(library%3D${ disableParms.name})`;
+        `?CRITERIA=(NAME%3D${disableParms.name})`;
```
davenice, line 110: this looks like it's being fixed - but I wonder how it was working before? do you know? have we got the right attribute name in the actual code?

codegen/templates/sdk/constants.hbs
BRSinha08: It was added there before the codegen existed (as part of the original hand-written constants file).

packages/sdk/__tests__/__unit__/disable/Disable.library.unit.test.ts
BRSinha08: Yes. CICS_LIBRARY_CRITERIA_FIELD = "NAME" in the constants is correct. The CMCI CICSLibrary resource uses NAME as its filter attribute (same as programs, transactions, etc.).The old test was a wrong .

packages/sdk/src/constants/CicsCmci.constants.ts
BRSinha08: this is existing code, we can change these to a uniform prefix.

PR #697, opened 2026-07-13
add informational badge

BRSinha08:
**What It Does**

**How to Test**

**Which Issue It Relates To**

**Review Checklist**
I certify that I have:
- [ ] tested my changes

**Additional Comments**


BRSinha08:
Proof of the task:

image

davenice:
I'm trying to catch up to where we are with this.

So our first PR was "send back the results, and an error notification."

Now - for some specific errors, like for instance the NOTPERMIT error, this is likely to be "business as usual" for a customer - they access their regions, for instance, and there are some they don't have access. No problem! They don't want an error notification every time as well as the results. We want to alert them in a softer way to the error. So we talked about some kind of tree decoration. Is that right?

So I think I'd be expecting that for some specific errors, instead of the error notification, we get the decoration on the tree and no error notification.

For errors that we haven't decided are "business as usual" - maybe we'd show both the error notification and the decoration? 🤔 I'm open to discussion on that idea!

I am not quite following the thread about cascading this up to the region level. The user opens a node in the tree. That node potentially doesn't include all results. So we decorate that node. Is there a reason why we'd decorate the parent regions? I think that could become confusing. Again - I'm happy to discuss. I couldn't manage to recreate that scenario in wiremock to feel it in action for myself.

Rajat-Sahrawat:
@chinmdas 
I think an additional icon at the region level is not necessary. I couldn't think of a strong use case, and it will not be very helpful since we cant show a specific error code if there are multiple.

BRSinha08:
> I'm trying to catch up to where we are with this.
> 
> So our first PR was "send back the results, and an error notification."
> 
> Now - for some specific errors, like for instance the NOTPERMIT error, this is likely to be "business as usual" for a customer - they access their regions, for instance, and there are some they don't have access. No problem! They don't want an error notification every time as well as the results. We want to alert them in a softer way to the error. So we talked about some kind of tree decoration. Is that right? :--> yes @davenice .
> 
> So I think I'd be expecting that for some specific errors, instead of the error notification, we get the decoration on the tree and no error notification. :---> @davenice I think we had decided for both, we will get error notification once, and also it will be visible from the tree as well.
> 
> For errors that we haven't decided are "business as usual" - maybe we'd show both the error notification and the decoration? 🤔 I'm open to discussion on that idea!
> 
> I am not quite following the thread about cascading this up to the region level. The user opens a node in the tree. That node potentially doesn't include all results. So we decorate that node. Is there a reason why we'd decorate the parent regions? I think that could become confusing. Again - I'm happy to discuss. I couldn't manage to recreate that scenario in wiremock to feel it in action for myself.

